PR #43, opened 2024-06-06
chore: publish docker image to repo's packages

fmaclen:


fmaclen:
Only had to update the settings on the original package, no need to publish a new one.

PR #70, opened 2024-07-07
feat: complete re-design, responsive UI & dark mode

fmaclen:
Closes #12 
Closes #14 

These designs are still a moving target but I've more or less settled with this layout:

v3 - Session - light
v3 - Session -dark
v3 - New session-1

new session mobile
no sessions mobile
session mobile

PR #75, opened 2024-07-10
feat: Add 'delete all knowledge' and its corresponding test

GregoMac1:


src/routes/+page.svelte
```diff
@@ -46,6 +46,13 @@
...
+	function deleteKnowledge(): void {
+		if (confirm('Are you sure you want to delete all knowledge?')) {
+			localStorage.removeItem('hollama-knowledge');
+			location.reload();
+		}
+	}
+
 	function deleteSettings(): void {
 		if (confirm('Are you sure you want to delete server settings?')) {
 			localStorage.removeItem('hollama-settings');
```
fmaclen, line 58: Couple of things:

- `deleteSessions()`, `deleteKnoledge()` and `deleteSettings()` are identical. Let's DRY them to a single `deleteStorage(item: string)` function. 
- We should probably add a dedicated type for `item`. In `store.ts` we are dynamically generating the `localStorage` keys. Let's add an enum and use that as the type, for example: `deleteStorage(item: StorageKey)`
- These would be the values of the `StorageKey` enum: `HOLLAMA_SETTINGS: 'hollama-settings', HOLLAMA_SESSIONS: 'hollama-sessions'` and so on...

tests/knowledge.test.ts
```diff
@@ -98,6 +98,46 @@ test('deletes knowledge', async ({ page }) => {
...
+					},
+					{
+						id: "m2jjac",
+						name: "Test 2",
+						content: "Hello World 2",
+						updatedAt: new Date().toISOString()
+					}
+				]
+			)
+		));
```
fmaclen, line 125: Minor thing, import [`MOCK_KNOWLEDGE`](https://github.com/fmaclen/hollama/blob/main/tests/utils.ts#L135-L148) and use that to seed the local storage.

One thing to keep in mind is that `page.evaluate()` runs _"inside the browser"_ so you won't have `MOCK_KNOWLEDGE` available, instead you need to pass it as an argument. Here's [an example](https://github.com/fmaclen/hollama/blob/main/tests/session.test.ts#L159-L183).

src/routes/+page.svelte
```diff
@@ -132,6 +139,7 @@
 		<Field name="danger-zone">
 			<svelte:fragment slot="label">Danger zone</svelte:fragment>
 			<Button variant="outline" on:click={deleteSessions}>Delete all sessions</Button>
+			<Button variant="outline" on:click={deleteKnowledge}>Delete all knowledge</Button>
```
fmaclen, line 142: Button's will end up looking something like this:

```suggestion
<Button variant="outline" on:click={()=> deleteStorage(StorageKey.HOLLAMA_KNOWLEDGE}>Delete all knowledge</Button>
```

fmaclen (changes requested):
Awesome start! 👍 

I think it'd be good to DRY things a bit now that we have a few consistent patterns. Left you some minor notes.

Also, let's remove the screenshots that were generated in `docs.test.ts`.
It's showing diffs for all them simply because your computer and mine have slightly different antialiasing. ~~I'll add a new issue~~ [here's an issue](https://github.com/fmaclen/hollama/issues/76) for generating the screenshots on demand so we don't get diffs on every PR, sorry about that 😬

GregoMac1:
> Awesome start! 👍
> 
> I think it'd be good to DRY things a bit now that we have a few consistent patterns. Left you some minor notes.
> 
> Also, let's remove the screenshots that were generated in `docs.test.ts`. It's showing diffs for all them simply because your computer and mine have slightly different antialiasing. ~I'll add a new issue~ [here's an issue](https://github.com/fmaclen/hollama/issues/76) for generating the screenshots on demand so we don't get diffs on every PR, sorry about that 😬

About the screenshots, you are right but we must also be aware that the Settings page's screenshot has changed because the "Delete all knowledge" button was added.

fmaclen:
> we must also be aware that the Settings page's screenshot has changed because the "Delete all knowledge" button was added.

Yeah, good point! Leave that one in.

---

Also, I just merged #70 and there is now a conflict with your `src/routes/+page.svelte`.

The **Settings** template is now in [`src/routes/settings/+page.svelte`](https://github.com/fmaclen/hollama/blob/main/src/routes/settings/%2Bpage.svelte), to settle the conflicts pull the latest from `main` and move your changes to that version of the file.

src/routes/+page.svelte
```diff
@@ -13,5 +13,4 @@
 			goto('/sessions');
 		}
 	});
-</script>
-
+</script>
```
fmaclen, line 16: Super minor thing, can you undo this change?

fmaclen (approved):
This looks great, thank you @GregoMac1!

PR #82, opened 2024-07-11
fix: Add timestamps to sessions and show dates using badges

### LINKED ISSUE #25: Add timestamps to sessions and sort the history by last updated
- Should add an `updatedAt` field to [`Sessions`](https://github.com/fmaclen/hollama/blob/main/src/lib/sessions.ts#L11-L17)
- Once Sessions have dates we should sort them in descending order, where the most recent ones are at the top.
- The current implementation [already does apply sorting](https://github.com/fmaclen/hollama/blob/main/src/routes/sessions/%2Blayout.svelte#L15-L16) but not based on the date.
- A similar implementation was done for [`Knowledge`](https://github.com/fmaclen/hollama/blob/main/src/lib/knowledge.ts#L5-L10)
- In the **Knowledge** section we should sort the list by `updatedAt` as well and also display it on the UI formatted with `date-fns` using the [`formatDistance`](https://date-fns.org/v3.6.0/docs/formatDistance) helper.
- The current Knowledge timestamp is rendered to the UI in [`src/routes/knowledge/+layout.svelte`](https://github.com/fmaclen/hollama/blob/main/src/routes/knowledge/%2Blayout.svelte#L20)

GregoMac1:
Closes #25


GregoMac1:
@fmaclen so far I've added the badges in the section and knowledge pages using the existing variants. The next thing is to apply the correct styles and to add the badges to the section/knowledge lists as well.

fmaclen:
@GregoMac1 here are the default styles for `<Badge>`:

```postcss
&--default {
  @apply bg-shade-3 text-muted;
}
```

And you'll have to update this too:
```typescript
export let variant: 'default' | 'positive' | 'warning' | undefined = 'default';
```

Couple of quick notes:
- Let's reduce the gap from `gap-x-3` to `gap-x-1` 
image
- I'm worried that the `about 1 hour ago` timestamps might be too long, does `date-fns` lets us do `1h` or `1w` and if older than 1 month then we just do `Dec 9, 2023`, similar to how Twitter/X does it?
- I haven't double check this but when I first tried your branch the "Sessions" view wouldn't load. I had to clear the sessions that in my `localStorage` and then it worked. This might be an issue with the `updatedAt` timestamp being missing, we need to account for that.

GregoMac1:
I tryed applying those styles and I got this result:
image
As you can see, the badges don't have borders. Should I add an extra style for that?

About the timestamps, I did some research and found [this](https://github.com/freearhey/dayjs-twitter) library than might help us. What do you think?

I think you're right about the sessions and the updatedAt field. I'll look further into it.

fmaclen:
@GregoMac1 sorry, I accidentally remove the `border` class.

```
&--default {
  @apply border-shade-3 text-muted;
}
```

fmaclen:
> About the timestamps, I did some research and found [this](https://github.com/freearhey/dayjs-twitter) library than might help us.

Ok, sounds good. Let's get rid of `date-fns` and use that one instead 👍 

Install the dependency as `devDependencies`: `npm install -D dayjs-twitter`.
SvelteKit has a build step and the external libraries get "bundled" so by default all external libraries should be `devDependencies`.

fmaclen:
> I think you're right about the sessions and the updatedAt field. I'll look further into it.

I think the easiest solution would be to make `updatedAt: string` an optional value.
And if it's `undefined` we don't show anything.

GregoMac1:
Hay @fmaclen, I've fixed all those details. Instead of using `dayjs-twitter`, I implemented a function to keep using `date-fns` and get a shorter format. I thought it would be easier to customise the format as we desire in the future.
Feel free to tell me if you'd like something to be changed!

src/lib/utils.ts
```diff
@@ -1,7 +1,29 @@
...
+		.replace(' hour ago', 'h')
+		.replace(' hours ago', 'h')
+		.replace(' day ago', 'd')
+		.replace(' days ago', 'd')
+		.replace(' month ago', 'mo')
+		.replace(' months ago', 'mo')
+		.replace(' year ago', 'y')
+		.replace(' years ago', 'y');
+	return shortFormat;
+}
```
fmaclen, line 29: I think it'd be easier to use your original library suggestion.
The are some edge-cases not covered by this function, for example:

image

I usually don't dislike the idea of writing our own _"simple helper functions"_ (if we can avoid using a 3rd-party library) but in this case I think it's likely there are other edge-cases that we haven't seen yet, so it's best to stick to a solution that's more mature.

Thankfully, we don't really need to do complex date manipulation so we can probably skip on `date-fns` for now.

Also, eventually I'd like to localize these strings but we'll cross that bridge when we have to.

src/routes/sessions/[id]/+page.svelte
```diff
@@ -169,12 +172,24 @@
...
+				{#if session.updatedAt}
+					<Badge variant="default" capitalize={false}>
+						{formatShortDistanceToNow(new Date(session.updatedAt), { addSuffix: true })}
+					</Badge>
+				{/if}
+				<Badge variant="default" capitalize={false}>
+					<p data-testid="model-name">{session.model}</p>
+				</Badge>
+			{/if}
+		</div>
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:192]: Not a huge fan of how this turned out visually (not your fault).
I need to play with the design a bit more.

In the mean time, let's revert back to the old `<p>` component:

```svelte
<p data-testid="model-name" class="text-sm text-muted">
   {isNewSession ? 'New session' : session.model}
</p>
```

src/routes/sessions/+layout.svelte
```diff
@@ -13,14 +14,20 @@
 	<svelte:fragment slot="list-items">
 		{#if $sessionsStore && $sessionsStore.length > 0}
 			<!-- Using slice() to reverse $sessionsStore without affecting the original array -->
-			{#each $sessionsStore.slice().reverse() as session (session.id)}
+			{#each $sessionsStore.slice().sort((a, b) => {
+				if (!a.updatedAt || !b.updatedAt) return 0;
+				return a.updatedAt < b.updatedAt ? 1 : -1;
+			}) as session}
```
fmaclen, line 20: Now that `Sessions` and `Knowledge` have an `updatedAt` field we could sort the store every time we update it:
- `saveKnowledge()`
- `saveSession()`

Something like this maybe...
```typescript
export const saveKnowledge = (knowledge: Knowledge): void => {
  // Retrieve the current knowledges
  const currentKnowledges = get(knowledgeStore) || [];

  // Find the index of the knowledge with the same id, if it exists
  const existingIndex = currentKnowledges.findIndex(k => k.id === knowledge.id);

  if (existingIndex !== -1) {
    // Update the existing knowledge
    currentKnowledges[existingIndex] = knowledge;
  } else {
    // Add the new knowledge if it doesn't exist
    currentKnowledges.push(knowledge);
  }

  // Sort the knowledges by updatedAt in descending order (most recent first)
  const sortedKnowledges = currentKnowledges.sort((a, b) => 
    new Date(b.updatedAt).getTime() - new Date(a.updatedAt).getTime()
  );

  // Update the store with the sorted knowledges
  knowledgeStore.set(sortedKnowledges);
};
```

And then we don't need to `.slice().sort()` on `$sessionStore` or `$knowledgeStore` .

src/routes/sessions/+layout.svelte
```diff
@@ -13,14 +14,20 @@
...
 				{@const knowledgeName = session.messages[0].knowledge?.name}
 				<SectionListItem
 					sitemap={Sitemap.SESSIONS}
 					id={session.id}
 					title={hasKnowledge ? session.messages[1].content : session.messages[0].content}
-					subtitle={knowledgeName ? `${session.model} — ${knowledgeName}` : session.model}
+					subtitle={(knowledgeName ? `${session.model} — ${knowledgeName}` : session.model) +
+						(session.updatedAt
+							? ` - ${formatShortDistanceToNow(new Date(session.updatedAt), { addSuffix: true })}`
+							: '')}
```
fmaclen, line 30: Hmm this is getting too complex, let's get rid of the string interpolated ` - `'s.

Instead I think we can do something like:
```svelte
{@const subtitles = [(knowledgeName ? session.model : knowledgeName), ... ]}

<SectoinListItem ... subtitle={subtitles.join(' - ')
```

src/routes/sessions/+layout.svelte
fmaclen: I know we talked about potentially DRY'n up these store files but for know let's only abstract the sorting logic. We can do the rest of the refactoring later.

src/routes/sessions/+layout.svelte
fmaclen: We also need to add this logic to [`Knowledge`](https://github.com/fmaclen/hollama/blob/main/src/routes/knowledge/%2Blayout.svelte#L16-L21)

fmaclen (changes requested):
Thank you for settling the conflicts!

I've played with these latest changes a bit more and ran into some edge-cases.
There's also some potential code/syntax optimizations we can make.

src/routes/sessions/+layout.svelte
GregoMac1: Great! I've already pushed this change (I've used `•` instead of `-` because I think it looks nicer, let me know what you think). Maybe in the future we could add these subtitles as badges, or leave the timestamp at the right of the title (in the same row) and use badges for the model and the knowledge.

fmaclen: Gotcha, `•` does look a bit nicer 👍 

In the near future want to re-think how these details on the sidebar and on the headers work, I'm not 100% convinced that including all of that metadata in those places is actually useful. Knowledge, for example, is already included in the history of the session so it probably doesn't make sense to include it there.

But until then I just want things not to _look broken_ 😅

src/lib/utils.ts
GregoMac1: I agree!

I have tried installing `dayjs-twitter` but I couldn't make it work, I think it's not made to work with typescript. I have searched for other libraries but I didn't find any.

In the mean time, I reverted it to the `date-fns` format which is too long but at least we have the edge cases covered.

Do you have any other ideas? At this point I don't think it would be a bad idea to write a custom function which covers all the edge cases, unless we find another library that does that for us.

fmaclen: No problem, `date-fns` with the default `formatDistanceToNow()` will work 👍

src/routes/knowledge/+layout.svelte
```diff
@@ -13,11 +14,14 @@
...
+					formatDistanceToNow(new Date(knowledge.updatedAt), { addSuffix: true })
+				]}
 				<SectionListItem
 					sitemap={Sitemap.KNOWLEDGE}
 					id={knowledge.id}
 					title={knowledge.name}
-					subtitle={knowledge.updatedAt}
+					subtitle={subtitles.join(' • ')}
 				/>
 			{/each}
```
fmaclen, line 26: In general I try to avoid having _single-use abstractions_, since the `subtitle` it's only going to be a timestamp for now we can just pass it directly and avoid setting a `{@const subtitles}`:

```suggestion
			{#each $knowledgeStore as knowledge}
				<SectionListItem
					sitemap={Sitemap.KNOWLEDGE}
					id={knowledge.id}
					title={knowledge.name}
					subtitle={formatDistanceToNow(new Date(knowledge.updatedAt), { addSuffix: true })}
				/>
			{/each}
```

src/lib/utils.ts
fmaclen, line 29: We should remove unused code.

src/routes/knowledge/[id]/+page.svelte
```diff
@@ -54,11 +55,18 @@
...
+			New knowledge
 			{:else}
-				{knowledge.updatedAt}
-			{/if}
+			<Badge variant="default" capitalize={false}>
+				<p data-testid="knowledge-timestamp">
+					{formatShortDistanceToNow(new Date(knowledge.updatedAt), { addSuffix: true })}
+					</p>
+					</Badge>
+					{/if} -->
```
Comment (fmaclen) [inline src/routes/knowledge/[id]/+page.svelte:66]: We should remove the commented code.

src/lib/knowledge.ts
```diff
@@ -1,12 +1,11 @@
...
-import { generateStorageId } from "./utils";
+import { get } from 'svelte/store';
+import { knowledgeStore } from '$lib/store';
 
 export interface Knowledge {
 	id: string;
 	name: string;
 	content: string;
-  updatedAt: string;
+	updatedAt: string;
```
fmaclen, line 8: `updatedAt` needs to be a conditional variable, otherwise it breaks the **Knowledge** section for users that created knowledge before this PR is merged.

```suggestion
	updatedAt?: string;
```

src/lib/knowledge.ts
```diff
@@ -29,21 +28,26 @@ export const loadKnowledge = (id: string): Knowledge => {
...
 
-	if (existingKnowledge !== -1) {
+	if (existingIndex !== -1) {
 		// Update the existing knowledge
-		currentKnowledges[existingKnowledge] = Knowledge;
+		currentKnowledges[existingIndex] = knowledge;
 	} else {
 		// Add the new knowledge if it doesn't exist
-		currentKnowledges.push(Knowledge);
+		currentKnowledges.push(knowledge);
```
fmaclen, line 43: Thanks for cleaning up these definitions, the variable names make more sense now 👍

src/lib/sessions.ts
```diff
@@ -44,16 +45,22 @@ export const saveSession = (session: Session): void => {
...
 		currentSessions.push(session);
 	}
 
-	// Update the store, which will trigger the localStorage update
-	sessionsStore.set(currentSessions);
+	// Sort the sessions by updatedAt in descending order (most recent first)
+	const sortedSessions = currentSessions.sort((a, b) => {
+		if (!a.updatedAt || !b.updatedAt) return 0;
+		return new Date(b.updatedAt).getTime() - new Date(a.updatedAt).getTime();
+	});
```
fmaclen, line 62: This _mostly works_ but there's an edge-case bug I found where if `a.updatedAt` exists and `b.updatedAt` is `undefined` then the guard clause on line 60 returns `0`, but that's incorrect because that condition should be sorted.

Also, the guard clause is missing in `saveKnowledge()`, but because those sorting functions will be identical I think it might be time to DRY these save functions altogether.

Can you try adding a new function on `src/lib/store.ts` called something like `function saveAndSortStore()` that abstracts the logic in `saveKnowledge()` and `saveSession()` which is entirely identical at this point.

This will be a good blueprint to eventually DRY'ing `loadKnowledge()` and `loadSession()` but we'll punt that for later.

src/routes/knowledge/[id]/+page.svelte
```diff
@@ -54,11 +55,18 @@
...
+			{isNewKnowledge
+				? 'New knowledge'
+				: formatDistanceToNow(new Date(knowledge.updatedAt), { addSuffix: true })}
```
Comment (fmaclen) [inline src/routes/knowledge/[id]/+page.svelte:69]: We now have several instances of `formatDistanceToNow(...)` so let's add a new function in `src/lib/utils.ts` called something like `export function formatTimestamp(date: string)` and replace all other instances with this new function.

src/routes/sessions/[id]/+page.svelte
```diff
@@ -169,12 +170,27 @@
...
+						{formatShortDistanceToNow(new Date(session.updatedAt), { addSuffix: true })}
+					</Badge>
+				{/if}
+				<Badge variant="default" capitalize={false}>
+					<p data-testid="model-name">{session.model}</p>
+				</Badge>
+			{/if} -->
+			<p data-testid="model-name" class="text-sm text-muted">
+				{isNewSession ? 'New session' : session.model}
+			</p>
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:192]: This is now only showing the model but not the timestamp.

I think ideally the `<Header>` metadata should be identical to the one on the sidebar so we can probably add a new function `export function formatSessionMetadata(session: Session)` in `src/lib/sessions.ts` that has all of this logic:

```svelte
{@const subtitles = [
	session.model,
	...(knowledgeName ? [knowledgeName] : []),
	...(session.updatedAt
		? [formatDistanceToNow(new Date(session.updatedAt), { addSuffix: true })]
		: [])
]}
```

And since we don't have to use the funky Svelte syntax we can do something a bit more readable like:

```typescript
export function formatSessionMetadata(session: Session)
    const subtitles = [session.model];
    if (knowledgeName) subtitles.push(knowledgeName);
    if (session.updatedAt) subtitles.push(formatTimestamp(session.updatedAt));
    return subtitles.join(' • ')
)
```

fmaclen (changes requested):
Alright, we are getting there 👍 

Most of the changes I'm suggesting are related to code duplication, and potential logic we can abstract.

There's still a minor bug in **Knowledge** when the user has items that don't have `updatedAt` so we need to make that conditional, similar issue we had with **Sessions**.

And also, make sure to delete all of the unused/commented code.

src/lib/knowledge.ts
GregoMac1: I agree, but Knowledge's field `updatedAt` was not added in this PR. I checked at the history of commits and that field was always mandatory for Knowledge.

GregoMac1:
Hey @fmaclen!

I ran into a bug with the `knowledge` field of a `session`. When I tried creating a new `session`, the field `knowledge` was not saved. I made a minor change to fix it.

I applied your corrections. The only thing I think needs to be fixed is that when a `session` has a `model`, an `updatedAt` and a `knowledge`, the concatenated string is not wraped in the sidebar:
image

src/lib/sessions.ts
GregoMac1: About this, I've been thinking: currently, the saveSession and saveKnowledge functions receive one param, the updated item.

An idea was to implement the function with two params, the store to update and the updated item (of generic type T).

The other idea was to keep the only one param (the item) and decide which store to update based in some attribute of the item, but I'm not so sure that this is a good idea.

What do you think?

Also, maybe we could extract this DRY to a separated issue and close this one for now.

fmaclen: Yeah, good point.

Let's just DRY the sorting function and keep everything else the same and can optimize the rest later.

src/lib/sessions.ts
```diff
@@ -44,16 +46,33 @@ export const saveSession = (session: Session): void => {
...
+	};
+	if (session.updatedAt) metadata.updatedAt = formatTimestampToNow(session.updatedAt);
+	if (session.knowledge?.name) metadata.knowledge = session.knowledge?.name;
+	return metadata;
+}
+
+export function formatSessionMetadata(session: Session) {
+	const subtitles: string[] = Object.values(getSessionMetadata(session));
+	return subtitles.join(' • ');
+}
```
fmaclen, line 78: In cases like this I'd like to avoid having single-use abstractions.
If we are not calling `getSessionMetadata()` anywhere else move it's logic inside `formatSessionMetadata()`.

fmaclen (approved):
Ok, I think we are there.

Left one final note about combining the logic of `getSessionMetadata` and `formatSessionMetadata`:
https://github.com/fmaclen/hollama/pull/82/files#r1681858158

Once that's done, feel free to merge away! 👍

src/lib/sessions.ts
GregoMac1: You're right. I had done that for a previous version of the files and then I forgot to revert it to a single function.

Issue #49, opened 2024-06-30
Reset session when an error occurs


fmaclen:
- Should probably need a dedicated UI component for handling error messages, such as a flash alert/toast
- When the prompt fails undo the last prompt into `prompt` field
- Should implement [these tests](https://github.com/fmaclen/hollama/blob/main/tests/session.test.ts#L268-L274).

![image](https://github.com/fmaclen/hollama/assets/1434675/f17ec3c6-ed7d-4722-b448-0950ca5c0682)

PR #99, opened 2024-07-18
fix: Disable Run button if server is disconnected or no model is chosen

### LINKED ISSUE #21: "Run" button should be disabled if the server is disconnected, no model is chosen or has no prompt
- The existing button is located in [`sessions/[id]/+page.svelte`](https://github.com/fmaclen/hollama/blob/main/src/routes/sessions/%5Bid%5D/%2Bpage.svelte#L196)
- Right now it's only disabled if it has `!prompt`
- The only [server check we do](https://github.com/fmaclen/hollama/blob/main/src/routes/%2Bpage.svelte#L13-L40) is in the **Settings** view and the server status not globally available.
- This is the current Typescript definition of [`Settings`](https://github.com/fmaclen/hollama/blob/main/src/lib/store.ts#L30-L34).
- We could probably save the server status to the `settingsStore` so we can access it from the **Sessions** view.
- Currently, we only check the status of the server whenever [`ollamaServer`](https://github.com/fmaclen/hollama/blob/main/src/routes/%2Bpage.svelte#L82C1-L83C1) changes. 
  - That means that if the server went _offline_ after starting a new session the status would still appear as _online_.
  - This is not ideal because submitting the form to an _offline_ server would cause an error but we should handle that behavior in #49 
- Should add a new test for this behavior in [`sessions.test.ts`](https://github.com/fmaclen/hollama/blob/main/tests/session.test.ts)

### LINKED ISSUE #80: When visiting `/`, redirect to "Settings" only when `ollamaServer` is offline
Blocked by #21
REF #70

```typescript
	beforeUpdate(() => {
		if (!$settingsStore?.ollamaServer) {
			// FIXME:
			// This should only rediect to /settings if the server is not connected
			goto('/settings');
		} else {
			goto('/sessions');
...
```

### LINKED ISSUE #105: Update list of available models when creating a "New Session"

GregoMac1:
Closes #21 
Closes #80 
Closes #105


GregoMac1:
@fmaclen the base functionality is done. I still have to add the tests.
About the error handling (when user tries to send a message and the server has just disconnected), how'd you like it to be shown?

fmaclen:
@GregoMac1 great 👍 

> About the error handling (when user tries to send a message and the server has just disconnected), how'd you like it to be shown?

I pushed a fix for #49 earlier today: https://github.com/fmaclen/hollama/pull/98

But there is a related issue to this feature that maybe you can do as part of this PR: https://github.com/fmaclen/hollama/issues/80

GregoMac1:
@fmaclen what do you think of closing this PR and continue working on #80 in a new PR?

src/routes/sessions/[id]/+page.svelte
```diff
@@ -47,6 +47,8 @@
 
 	const shouldConfirmDeletion = writable(false);
 
+	$: isServerConnected = !!$settingsStore?.ollamaServer;
+	$: isModelChosen = !!$settingsStore?.ollamaModel;
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:51]: > what do you think of closing this PR

@GregoMac1 unless you mean closing it without merging it, this PR it needs additional work.

If `$settingsStore.ollamaServer` is defined that doesn't necessarily mean that `isServerConnected == true`.

`isServerConnected` will be `true` if we are able to get a status 200 when fetching models from `/api/tags`.
Right now that only happens when we call `getModelsList()` in `settings/+page.svelte`.

If you want to limit this PR to disabling the submit button if there are no models and do the rest in another one, that's totally fine.

But in that case I'd also avoid adding a new `isModelChosen` variable, let's just set that directly in the `<ButtonSubmit>` component without adding single-use abstractions; this syntax will still be reactive:
```
disabled={!prompt || !$settingsStore?.ollamaModel}
```

And also, whatever we do here will need a test assertion/s.

GregoMac1:
@fmaclen I improved this. Now, it checks for `/api/tags` every time the user selects a different session.

There is a problem with the way the errors are stored. Steps to reproduce:
- `ollama serve`.
- Choose some session.
- Turn down ollama.
- Choose another session or New session (now system error message is displayed and prompt field is disabled).
- `ollama serve` again.
- Choose another session. Here, the prompt field is enabled again but the system error message is still displayed. The only way to get rid of it is refreshing the page.

As far as I understand, this happens because the system error message is stored within the session's messages. When I refresh the page, the messages are fetched from the browser's localStorage again.

I think we could separate the error from the session messages, or find a way to be able to restore the original messages without having to refresh the page.

src/routes/sessions/[id]/+page.svelte
```diff
@@ -54,6 +55,23 @@
 	$: if ($settingsStore?.ollamaModel) session.model = $settingsStore.ollamaModel;
 	$: knowledge = knowledgeId ? loadKnowledge(knowledgeId) : null;
 	$: shouldFocusTextarea = !isPromptFullscreen;
+	$: {
+		if (session && session.id !== $currentSessionId) {
+			getModelsList();
+			currentSessionId.set(session.id);
+		}
+	}
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:63]:
```suggestion
	$: {
		if (session?.id !== $currentSessionId) {
			getModelsList();
			$currentSessionId = session.id;
		}
	}
```

src/routes/sessions/[id]/+page.svelte
```diff
@@ -54,6 +55,23 @@
...
+	async function getModelsList() {
+		if (!$settingsStore) return;
+		try {
+			$settingsStore.ollamaModels = (await ollamaTags()).models;
+			//TODO hide error message if present
+		} catch {
+			$settingsStore.ollamaModels = [];
+			handleError(new Error('Failed to fetch'));
+		}
+	}
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:74]: I don't think we want to call `handleError()` here, it causes a bug when creating a **New session** and Ollama is offline it renders the error in the as a session message, that triggers the the model drop-down and knowledge fields to disappear (as if it's no longer a "new session").

Instead I'm thinking we could do something like this:
```svelte
<svelte:fragment slot="label">
	Prompt

	{#if !$settingsStore?.ollamaModels.length}
		<a class="ml-auto" href="/settings">
			<Badge variant="warning">disconnected</Badge>
		</a>
	{/if}
</svelte:fragment>
```

Which looks like this:

localhost_5173_sessions_y7ozps

TBH, I don't love this solution but until we have a toast/flash alert system this would probably do.

src/routes/sessions/[id]/+page.svelte
```diff
@@ -260,7 +278,13 @@
 						{/key}
 
 						<div class="flex w-full">
-							<ButtonSubmit {handleSubmit} hasMetaKey={isPromptFullscreen} disabled={!prompt}>
+							<ButtonSubmit
+								{handleSubmit}
+								hasMetaKey={isPromptFullscreen}
+								disabled={!prompt ||
+									!$settingsStore?.ollamaModels.length ||
+									!$settingsStore?.ollamaModel}
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:286]: Do you know if there could there be a situation where `$settingsStore.ollamaModel` is defined but `$settingsStore?.ollamaModels.length == []` isn't?

src/routes/settings/+page.svelte
```diff
@@ -140,27 +136,17 @@
...
+				<Button variant="link" href="https://github.com/fmaclen/hollama" target="_blank">
 					Github
 				</Button>
 			</p>
 			<p class="p">
 				Made by
-				<Button variant="link" href="https://fernando.is" target="_blank">
-					@fmaclen
-				</Button>
+				<Button variant="link" href="https://fernando.is" target="_blank">@fmaclen</Button>
```
fmaclen, line 149: I feel like most of these formatting changes are adding too much noise the PRs; and I probably it's mostly on me because I don't format on save 😬 

I'm thinking it might be good run a linter check in CI.

src/lib/ollama.ts
```diff
@@ -42,6 +42,16 @@ export type OllamaTagResponse = {
...
 
+export async function ollamaTags() {
+	const settings = get(settingsStore);
+	if (!settings) throw new Error('No Ollama server specified');
+
+	const response = await fetch(`${settings.ollamaServer}/api/tags`);
+	if (!response.ok) throw new Error('Failed to fetch Ollama tags');
+
+	return response.json() as Promise<OllamaTagResponse>;
+}
```
fmaclen, line 53: No changes needed here, but just so you know I will probably end up replacing all of these `fetch()` calls with Ollama's own JS library: https://github.com/ollama/ollama-js as part of https://github.com/fmaclen/hollama/issues/103#issuecomment-2253706258

fmaclen (changes requested):
So far so good! I think we are almost there with this one.

Also, I don't think we need to add a brand new test for this logic but see if you can add a couple of assertions for the "offline state" to an existing test.

Here's a reference syntax that simulates network and other errors that might be useful:
https://github.com/fmaclen/hollama/blob/main/tests/session.test.ts#L386-L418

src/routes/settings/+page.svelte
fmaclen: https://github.com/fmaclen/hollama/issues/121

fmaclen: I have a PR ready for this: https://github.com/fmaclen/hollama/pull/122

I'll wait to merge it after your other 2 PRs are merged first so I'll handle any potential conflicts.

src/routes/sessions/[id]/+page.svelte
Comment (GregoMac1) [reply src/routes/sessions/[id]/+page.svelte:293]: Sorry for the delay.

Currently, if you are connected to the server, then choose a model, and then disconnect, `ollamaModels` will be reseted to `[]` when you change the session, but `ollamaModel` will still be defined as your previous chosen model.

I think this behavior is correct because if I have my model chosen and I get disconnected from the server, I'd like to have the same model chosen when I get connected again. What do you think?

GregoMac1:
I ended up adding a new test, I thought I would be cleaner and easier to understand and debug.

src/routes/sessions/[id]/+page.svelte
Comment (fmaclen) [reply src/routes/sessions/[id]/+page.svelte:293]: Oh, right. Good point!

fmaclen (changes requested):
Alright, I think this PR is ready to merge 👍
But there are a few conflicts against `main` that need to be settled.

PR #101, opened 2024-07-19
fix: pass knowledge as system prompt

### LINKED ISSUE #100: Knowledge as system prompt
Currently, knowledge is passed as xml block before the first user input.
If it's not only information but instructions for the model (you are ..., pretend to be ...), it gets confused and interprets it as if I – the user – was doing these things.
When using the ollama cli and setting the knowledge as system prompt, the model behaves as expected.

The knowledge system itself is great, especially if later multiple knowledges can be passed to the model.
I'd suggest adding the possibility to set (one of) the knowledge(s) as system prompt.

fmaclen:
Closes #100


fmaclen:
Will be implemented in https://github.com/fmaclen/hollama/pull/106

PR #107, opened 2024-07-20
fix: save user theme to settingsStore

### LINKED ISSUE #78: Save user preference of "color theme" to localStorage
REF #70 

We could probably add a new `userTheme: 'dark' | 'light;` to [`$settingsStore`](https://github.com/fmaclen/hollama/blob/main/src/lib/store.ts#L30-L36)

GregoMac1:
Closes #78 
~~TODO: The theme is correctly stored but, if the saved theme is dark, when user reloads the page the initial theme is light and then it turns to dark (because it gets changed onMount).~~

### DIFF
--- a/src/lib/store.ts
+++ b/src/lib/store.ts
@@ -44,7 +44,7 @@ export interface Settings {
 	ollamaServer: string | null;
 	ollamaModel: string | null;
 	ollamaModels: OllamaModel[];
-	userTheme: 'light' | 'dark' | undefined;
+	userTheme?: 'light' | 'dark';
 }
 
 export enum StorageKey {
--- a/src/lib/store.ts
+++ b/src/lib/store.ts
@@ -44,6 +44,7 @@ export interface Settings {
 	ollamaServer: string | null;
 	ollamaModel: string | null;
 	ollamaModels: OllamaModel[];
+	userTheme: 'light' | 'dark' | undefined;
 }
 
 export enum StorageKey {
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -18,24 +18,16 @@
 	$: theme = $settingsStore?.userTheme;
 
 	onMount(() => {
-		if (browser && !theme) {
-			const systemTheme = window.matchMedia('(prefers-color-scheme: dark)').matches
-				? 'dark'
-				: 'light';
-			settingsStore.update((store) => ({
-				...store!,
-				userTheme: systemTheme
-			}));
-		}
+		if (!$settingsStore || !browser || theme) return;
+		$settingsStore.userTheme = window.matchMedia('(prefers-color-scheme: dark)').matches
+			? 'dark'
+			: 'light';
 	});
 
 	function toggleTheme() {
 		theme = theme === 'light' ? 'dark' : 'light';
 		document.documentElement.setAttribute('data-color-theme', theme);
-		settingsStore.update((store) => ({
-			...store!,
-			userTheme: theme
-		}));
+		if ($settingsStore) $settingsStore.userTheme = theme;
 	}
 </script>
 
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-	import { onMount } from 'svelte';
 	import { env } from '$env/dynamic/public';
 	import { page } from '$app/stores';
 	import { Brain, MessageSquareText, Settings2, Sun, Moon } from 'lucide-svelte';
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -22,20 +22,20 @@
 			const systemTheme = window.matchMedia('(prefers-color-scheme: dark)').matches
 				? 'dark'
 				: 'light';
-			settingsStore.update((store) => {
-				store!.userTheme = systemTheme;
-				return store;
-			});
+			settingsStore.update((store) => ({
+				...store!,
+				userTheme: systemTheme
+			}));
 		}
 	});
 
 	function toggleTheme() {
 		theme = theme === 'light' ? 'dark' : 'light';
 		document.documentElement.setAttribute('data-color-theme', theme);
-		settingsStore.update((store) => {
-			store!.userTheme = theme;
-			return store;
-		});
+		settingsStore.update((store) => ({
+			...store!,
+			userTheme: theme
+		}));
 	}
 </script>
 
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -5,6 +5,7 @@
 	import { Brain, MessageSquareText, Settings2, Sun, Moon } from 'lucide-svelte';
 
 	import '../app.pcss';
+	import { settingsStore } from '$lib/store';
 
 	$: pathname = $page.url.pathname;
 	const SITEMAP = [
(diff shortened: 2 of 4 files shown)


src/routes/+layout.svelte
```diff
@@ -13,17 +14,23 @@
...
-		// Set initial theme based on user's browser preference
-		const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
-		theme = prefersDark ? 'dark' : 'light';
+		if (!theme) {
+			// Set initial theme based on user's browser preference
+			const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
+			theme = prefersDark ? 'dark' : 'light';
+		}
 		document.documentElement.setAttribute('data-color-theme', theme);
 	});
```
fmaclen, line 35: > when user reloads the page the initial theme is light and then it turns to dark (because it gets changed onMount)

I think the correct way of doing this is not to use `onMount()` and instead use: [`import { browser } from '$app/environment';`](https://kit.svelte.dev/docs/modules#$app-environment-browser)

Here's an example ([Source](https://dev.to/willkre/persistent-theme-switch-dark-mode-with-svelte-sveltekit-tailwind-1b9g)):

```svelte
<script lang="ts">
    import { browser } from '$app/environment';

    let darkMode = true;

    function handleSwitchDarkMode() {
        darkMode = !darkMode;

        localStorage.setItem('theme', darkMode ? 'dark' : 'light');

        darkMode
            ? document.documentElement.classList.add('dark')
            : document.documentElement.classList.remove('dark');
    }

    if (browser) {
        if (
            localStorage.theme === 'dark' ||
            (!('theme' in localStorage) && window.matchMedia('(prefers-color-scheme: dark)').matches)
        ) {
            document.documentElement.classList.add('dark');
            darkMode = true;
        } else {
            document.documentElement.classList.remove('dark');
            darkMode = false;
        }
    }
</script>
```

In another project I had a similar implementation that used this technique:
https://github.com/search?q=repo%3Afmaclen%2Fcanutin-desktop%20colorThemeStore&type=code

But that alone won't solve the issue with the FOUC (_Flash of Unstyled Content_), there is one more piece to the puzzle in [`app.html`](https://github.com/fmaclen/canutin-desktop/blob/453d1869c0edbe3a707b57c7645d4bedc80ed90e/sveltekit/src/app.html#L9-L16) where we read the `localStorage` outside of Svelte.

You can try it out in the live demo: https://demo.canutin.com/settings

fmaclen (changes requested):
This looks good so far 👍 

I'd like to get that initial theme change bug you pointed out too.
I left you some suggestions on how to fix that issue.

src/lib/store.ts
```diff
@@ -44,6 +44,7 @@ export interface Settings {
 	ollamaServer: string | null;
 	ollamaModel: string | null;
 	ollamaModels: OllamaModel[];
+	userTheme: 'light' | 'dark' | undefined;
```
fmaclen, line 47: I just noticed that here you can do:
```suggestion
	userTheme?: 'light' | 'dark';
```

src/routes/+layout.svelte
GregoMac1: I tried these options and I came to a solution where the initial theme is selected in [`app.html`](https://github.com/fmaclen/hollama/blob/20b39847bfc324c884a15f4ccd2ea36a147500e5/src/app.html#L11-L20). Then, [`+layout.svelte`](https://github.com/fmaclen/hollama/blob/20b39847bfc324c884a15f4ccd2ea36a147500e5/src/routes/%2Blayout.svelte#L18-L25) only handles the theme switch. This removes the FOUC and works correctly.

src/routes/+layout.svelte
```diff
@@ -13,19 +13,16 @@
...
+			return store;
+		});
 	}
-
-	onMount(() => {
-		// Set initial theme based on user's browser preference
-		const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
-		theme = prefersDark ? 'dark' : 'light';
-		document.documentElement.setAttribute('data-color-theme', theme);
-	});
```
fmaclen, line 28: I'm not sure if you finished working on this PR but this `prefersDark` was needed to determine what the default color theme should have been.

My browser is set to _"prefers light mode"_, but in [the latest version of this PR](https://f17e6891.hollama.pages.dev/) the default option is to "switch to light mode":

image

GregoMac1: > I'm not sure if you finished working on this PR but this `prefersDark` was needed to determine what the default color theme should have been.

I moved this logic to `app.html`, I thought it should have worked the same way as before.


> My browser is set to _"prefers light mode"_, but in [the latest version of this PR](https://f17e6891.hollama.pages.dev/) the default option is to "switch to light mode":

That's weird. It behaves correctly when I try it. I'll look further into that.

fmaclen: > That's weird. It behaves correctly when I try it. I'll look further into that.

I'm able to consistently replicate the behavior when I delete the existing `localsStorage` `hollama-settings` and refresh the page.

fmaclen: Also, not sure if you know this but you can emulate the browser settings in Inspector with the "Run command" menu:

image

GregoMac1: Thanks for the advice!

The problem should be fixed now.

GregoMac1:
@fmaclen this has conflicts with the tests but the conflicts should be gone if the branches get merged. I tested it locally.

src/routes/+layout.svelte
```diff
@@ -13,19 +15,28 @@
...
+		}
+	});
 
 	function toggleTheme() {
 		theme = theme === 'light' ? 'dark' : 'light';
 		document.documentElement.setAttribute('data-color-theme', theme);
+		settingsStore.update((store) => ({
+			...store!,
+			userTheme: theme
+		}));
```
fmaclen, line 38: This is not necessarily wrong, but one thing to note is that inside a Svelte component you don't need to call `settingsStore.update()`, you can simply do:

```ts
$settingsStore!.userTheme = systemTheme;
```

> Assignments to $-prefixed variables require that the variable be a writable store, and will result in a call to the store's .set method. — https://svelte.dev/docs/svelte-components#script-4-prefix-stores-with-$-to-access-their-values

Also, I'm not sure what the best way to handle this is but if `$settingsStore == null` (which I don't know if it's even possible in this case), calling `$settingsStore!` would throw an error to the console, right? Which I guess it's useful because we are always expecting the store to be present not matter what.

[In other places](https://github.com/fmaclen/hollama/blob/main/src/routes/sessions/%5Bid%5D/%2Bpage.svelte#L59) I added a conditional check, but maybe we actually need to intentionally throw an error if we really are expecting `settingsStore` to be set at that point.

I personally almost never use the syntax `variable!`, feels a bit like a code smell. My gut tells me that in both `onMount()` and `toggleTheme()` we'd be better off adding a guard clause, specially `onMount()` since running the rest of the code in that section doesn't make sense if the store isn't present for whatever reason.

What do you think?

fmaclen (approved):
The functionality appears to be working fine now and the rest of the PR looks good 👍 
Just left you a note about a potential code smell when updating the store.

src/routes/+layout.svelte
fmaclen: Maybe something like this can make the code a bit more readable?

```ts
onMount(() => {
	if (!$settingsStore || !browser || theme) return;
	$settingsStore!.userTheme = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
});
```

```ts
function toggleTheme() {
	theme = theme === 'light' ? 'dark' : 'light';
	document.documentElement.setAttribute('data-color-theme', theme);
	if ($settingsStore) $settingsStore.userTheme = theme;
}
```

GregoMac1: Nice! I agree. I don't like using `variable!` either, but I also thought about the need of `settingsStore` to always be present.

I've just commited this, so I am merging now.

PR #108, opened 2024-07-21
fix: add page title

### LINKED ISSUE #74: Add relevant `<title>` tags
- Here's an example of a similar implementation I've done in [another Svelte project](https://github.com/fmaclen/canutin-desktop/blob/master/sveltekit/src/lib/components/Head.svelte).
- The format of the titles should be:
  -  `["Title of the session", "Sessions", "Hollama"]`
  -  `["Name of the knowledge", "Knowledge", "Hollama"]`
  - `["Settings", "Hollama"]`

GregoMac1:
Closes #74


fmaclen (changes requested):
This implementation appears to be more complex than what I had in mind.

In #74 I suggested simply adding a new component called [`<Head>`](https://github.com/fmaclen/canutin-desktop/blob/master/sveltekit/src/lib/components/Head.svelte) that can accept a `title: string[]`.

Then we include this `<Head>` component on every [`+page.svelte`](https://github.com/fmaclen/canutin-desktop/blob/master/sveltekit/src/routes/transaction/%5Bslug%5D/%2Bpage.svelte#L55) where we want a custom title.

For example, in `/routes/knowledge/[id]/+page.svelte` we could do:
```svelte
<Head title={[knowledge.name, 'Knowledge']} />
```

That way you can avoid dealing with `$settingsStore` and most of the other changes (except the implementation of `getSessionTitle()` which I think it's a good idea 👍 ).

Will also need to remove `<title>` from [`+layout.svelte`](https://github.com/fmaclen/hollama/blob/d0852cd7850c717cf4aa67f2e548b0d7f9874830/src/routes/%2Blayout.svelte#L41).

GregoMac1:
@fmaclen I applied the changes.

Now there's only one thing that I'd like you to help me with.

If you go to the Sessions page, the title will be `Sessions • Hollama`. If you enter a session, the title will be `sessionTitle • Sessions • Hollama`.

The problem appears when you go back to the Sessions page by clicking it in the sidebar, because the title remains on `sessionTitle • Sessions • Hollama` when it should change back to `Sessions • Hollama`. This behavior also happens with Knowledge.

How would you fix it?

fmaclen:
> The problem appears when you go back to the Sessions page by clicking it in the sidebar, because the title remains on sessionTitle • Sessions • Hollama

I think we should use the `<Head/>` component only in files named `+page.svelte`, not in `+layout.svelte`.

If that doesn't work, we will need to wrap the component in a logic block [`{#key foo}`](https://svelte.dev/docs/logic-blocks#key) which triggers a re-render (of whatever components you include the block) anytime `foo` changes.

GregoMac1:
You were right! That was the fix, thank you. I think this is ready to be closed now.

GregoMac1:
I've just merged `main` to this branch and fixed the conflicts. I hope i did it correctly 😅

src/routes/sessions/[id]/+page.svelte
```diff
@@ -52,14 +54,17 @@
...
 	$: isLastMessageFromUser = session?.messages[session.messages.length - 1]?.role === 'user';
 	$: session && scrollToBottom();
 	$: if ($settingsStore?.ollamaModel) session.model = $settingsStore.ollamaModel;
 	$: knowledge = knowledgeId ? loadKnowledge(knowledgeId) : null;
 	$: shouldFocusTextarea = !isPromptFullscreen;
 
+	afterNavigate(() => {
+		session = loadSession(data.id);
+	});
+
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:67]: Was there a particular reason for this change?
If not let's revert it and also revert the change on line 43.

src/routes/settings/+page.svelte
```diff
@@ -21,11 +22,12 @@
...
 	let ollamaTagResponse: OllamaTagResponse | null = null;
 
-	$: settingsStore.set({
+	$: settingsStore.update((settings) => ({
+		...settings,
 		ollamaServer,
 		ollamaModels: ollamaTagResponse?.models || [],
 		ollamaModel
-	});
+	}));
```
fmaclen, line 30: I think this may be a left-over change, if so let's also revert it.

fmaclen (changes requested):
Almost done, left you a couple of minor changes.

src/routes/+layout.svelte
```diff
@@ -38,7 +39,7 @@
 		></script>
 	{/if}
 
-	<title>Hollama</title>
+	<Head />
```
fmaclen, line 42: We should probably get rid of this one too.

src/routes/settings/+page.svelte
GregoMac1: You're right, but I think this change should be present in #107 because `settingsStore` will also have `userTheme`, so if you do `settingsStore.set` with only the ollama variables, the `userTheme` will get lost.

fmaclen: I see, if that's the case let's fix it in #107 and remove it from this PR, that way both are easier to review.

src/routes/sessions/[id]/+page.svelte
Comment (GregoMac1) [reply src/routes/sessions/[id]/+page.svelte]: There was a reason for the way I was handling the title before, but now it is not necessary anymore.

This way (using `afterNavigate`) is how knowledge is handled, though. However, I have already restored it to the previous version.

fmaclen (approved):
This looks good, let's ship it! 🚀

PR #118, opened 2024-07-26
fix: release version is displayed in the UI

fmaclen:
- Updates `package.json` with the version output of `semantic-release`
- Builds release for Cloudflare Pages and deploys it
- Builds Docker image and publishes it to GH's registry

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -74,7 +74,7 @@ jobs:
           image: hollama-gh-actions
           registry: ghcr.io
           multiPlatform: true
-          platform: linux/amd64,linux/arm64,linux/arm/v7
+          platform: linux/amd64,linux/arm64
           tags: latest, ${{ needs.release.outputs.new_release_version }}
           username: ${{ secrets.GH_REGISTRY_USERNAME }}
           password: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM node:20-alpine
 
 WORKDIR /app
 
-ENV ADAPTER 'docker-node'
+ENV ADAPTER='docker-node'
```
fmaclen, line 5: Fixes legacy syntax

.github/workflows/publish.yml
```diff
@@ -42,22 +41,40 @@ jobs:
...
-      - name: Login to GitHub Container Registry
-        uses: docker/login-action@v3
+    name: Docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Update app version in package.json
+        run: node .github/workflows/publish/update-package-json.cjs ${{ needs.release.outputs.new_release_version }}
```
fmaclen, line 69: Even though `semantic-release` does automatically update `package.json` with the latest version, we want `semantic-release` to remain as a stand-alone job so we don't have to re-run it in case the build jobs fail.

Therefore we need to manually update the version in `package.json` for the `cloudflare` and `docker` jobs, based on the output of the `release` job.

PR #119, opened 2024-07-27
fix: add auto scroll on sessions

### LINKED ISSUE #13: Toggable auto-scroll

GregoMac1:
Closes #13 
- As long as the AI model generates the reply, the screen scrolls automatically.
- If the user scrolls up, then the auto scroll stops.
- If the user scrolls to the bottom of the screen, it starts scrolling automatically again.


src/routes/sessions/[id]/+page.svelte
```diff
@@ -61,6 +62,14 @@
 	$: if ($settingsStore?.ollamaModel) session.model = $settingsStore.ollamaModel;
 	$: knowledge = knowledgeId ? loadKnowledge(knowledgeId) : null;
 	$: shouldFocusTextarea = !isPromptFullscreen;
+	$: if (messageWindow) {
+		messageWindow.addEventListener('scroll', handleScroll);
+	}
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:67]:
```suggestion
	$: if (messageWindow) messageWindow.addEventListener('scroll', handleScroll);
```

fmaclen (changes requested):
This works great 👍 
Only left you a minor syntax cleaning suggestion.

Also before we ship this, it'd be great to finally add a test for this feature.

I didn't have tests when I started the app, so over time I started adding `test.skip()`'s as a to-do list of functionality that wasn't tested and should have.

There is only one more `test.skip()` which is:
```typescript
// tests/session.test.ts

test.skip('auto-scrolls to the bottom when new messages are added', async ({ page }) => {
  // TODO
});
```

It should be a really simple test, I think you can just generate a mocked completion and `expect` the scroll height to be different before and after clicking "Run".

GregoMac1:
I added the test! Please check if we're ready to merge now.

tests/session.test.ts
```diff
@@ -442,8 +442,41 @@ test.describe('Session', () => {
...
+		await page.getByLabel('Model').selectOption(MOCK_API_TAGS_RESPONSE.models[0].name);
+		await promptTextarea.fill('Who would win in a fight between Emma Watson and Jessica Alba?');
+		await expect(runButton).toBeEnabled();
+
+		// Mock the response and submit the prompt
+		await mockCompletionResponse(page, MOCK_SESSION_1_RESPONSE_1);
+		await page.keyboard.press('Shift+Enter');
+		await expect(sessionMetadata).toHaveText('New session');
+
+		await page.keyboard.press('Enter');
```
fmaclen, line 466: You can skip all of the assertions and keyboard shortcuts, we have other tests that cover that functionality.
Let's just do:

```suggestion
		await mockCompletionResponse(page, MOCK_SESSION_1_RESPONSE_1);
		await page.getByText('Run').click();
```

tests/session.test.ts
```diff
@@ -442,8 +442,41 @@ test.describe('Session', () => {
...
+					'I am unable to provide subjective or speculative information, including fight outcomes between individuals.'
+			})
+		).toBeVisible();
+
+		// Check if the session history auto-scrolls to the bottom
+		const scrollHeight = await sessionHistory.evaluate((el) => el.scrollHeight);
+		const clientHeight = await sessionHistory.evaluate((el) => el.clientHeight);
+		const scrollTop = await sessionHistory.evaluate((el) => el.scrollTop);
+
+		expect(scrollTop + clientHeight).toBe(scrollHeight);
```
fmaclen, line 479: I don't think this is testing our scroll logic, we are just doing math on some random numbers 😅 

What we need to do is check that the `clientHeight` and `scrollHeight` were different before and after clicking "Run".

We are also missing assertions for the `userScrolledUp` behavior but I think this would be tricky to test because  the auto-scrolling with mocked completions is almost instantaneous and will likely lead to race-conditions.

Let's just add a comment for now that includes the variable name so it's easy to find in the future:
```
// TODO: Add assertions for the `userScrolledUp` behavior
```

fmaclen (changes requested):
Looks good, but we need to do a few minor changes to the tests.

Also, there are some unresolved conflicts on this branch, let me know if you need help settling them.

fmaclen (approved):
LGTM 🚀

This bug was irritating me so much :rage4:

PR #120, opened 2024-07-27
fix: sort the available model list alphabetically

### LINKED ISSUE #104: Order the available models drop-down
Currently, when picking a model on the new sessions page shows the models unordered.
Sorting my name or model file size would come handy when frequently switching between different models. This would allow to quickly locate the desired model.

~~Also, the list of available models does only get updated when opening the settings page which is a bit annoying.~~
https://github.com/fmaclen/hollama/issues/105

binarynoise:
Closes #104

- sort models received from ollama


fmaclen:
Thanks for the contribution!

There's currently [another PR](https://github.com/fmaclen/hollama/pull/99/files#diff-31ae89989234ded3c6f93e8e5fe7cc2c30e93536cc027383faae05989bf79205R45-R53) that's almost ready to be merged and that moved the `getModelList()` logic elsewhere, so this sorting function will also have to be moved.

Let's put this on hold until #99 is merged. I'll let you know when it is.

Also, I'd like to add some test assertions for the sorting function.
I think the simplest implementation would be to add a new entry in [`MOCK_API_TAGS_RESPONSE`](https://github.com/fmaclen/hollama/blob/main/tests/utils.ts#L5-L38), for example:

```js
models: [
    { ... },
    {
	name: "TheBloke/CodeLlama-70B-Python-AWQ",
	model: "TheBloke/CodeLlama-70B-Python-AWQ",
	modified_at: "2023-12-01T10:06:18.679361519-05:00",
	size: 4108928240,
	digest: "ca4cd4e8a562465d7cf0512fc4d4dff3407f07800b01c9a6ee9dd107001704b9",
	details: {
		parent_model: "",
		format: "gguf",
		family: "llama",
		families: null,
		parameter_size: "70B",
		quantization_level: "Q4_0"
	}
    }
]
```

And then in [`test("displays model list and updates settings store")`](https://github.com/fmaclen/hollama/blob/main/tests/settings.test.ts#L8-L22) we should _expect_ the order of the models to be sorted correctly:

```typescript
// Check if the model list contains the expected models
await expect(modelSelect).toContainText(MOCK_API_TAGS_RESPONSE.models[0].name);
await expect(modelSelect).toContainText(MOCK_API_TAGS_RESPONSE.models[1].name);
await expect(modelSelect).toContainText(MOCK_API_TAGS_RESPONSE.models[2].name);

await modelSelect.selectOption(MOCK_API_TAGS_RESPONSE.models[1].name);

// Check if the settings store is updated with the selected model
const localStorageValue = await page.evaluate(() => window.localStorage.getItem('hollama-settings'));
if (!localStorageValue) throw new Error('No local storage value');
const parsedLocalStorageValue = JSON.parse(localStorageValue);
expect(parsedLocalStorageValue.ollamaModel).toBe(MOCK_API_TAGS_RESPONSE.models[1].name);
// Check that the models are sorted alphabetically (excluding 3rd party repositories)
expect(parsedLocalStorageValue.ollamaModels[0].name).toBe(MOCK_API_TAGS_RESPONSE.models[2].name);
expect(parsedLocalStorageValue.ollamaModels[1].name).toBe(MOCK_API_TAGS_RESPONSE.models[0].name);
expect(parsedLocalStorageValue.ollamaModels[2].name).toBe(MOCK_API_TAGS_RESPONSE.models[1].name);
```

binarynoise:
:+1: I'll take another look when the other PR is merged. I would defer the testing until then too. Luckily ts isn't too crazy, so writing tests will probably work for me.

For me, having the path locally already is a great help (I currently have about 40 models lying around waiting to be decided on whether they work for me or don't).

fmaclen:
@binarynoise alright, I think you can proceed 👍 

The best place to sort the models now would be in [`ollamaTags()`](https://github.com/fmaclen/hollama/blob/main/src/lib/ollama.ts#L28-L31).

Also, any chance we can avoid using the non-null `!`'s?
As a reference, we have a similar [sorting function](https://github.com/fmaclen/hollama/blob/main/src/lib/store.ts#L28-L35) that has some guard clauses to avoid `!`'s.

binarynoise:
I reimplemented the sort.
I decided not to split the repo off, because in a private project I don't do that either.
That way, we don't have that `!` problem any more.

fmaclen:
@binarynoise thank you, this looks great!

Any chance you can update the test? I think you can mostly copy/paste my suggestion here:
https://github.com/fmaclen/hollama/pull/120#issuecomment-2254157126

binarynoise:
Here you go, I added the tests.

src/lib/ollama.ts
```diff
@@ -28,5 +9,18 @@ export async function ollamaTags() {
...
-	return response.json() as Promise<OllamaTagResponse>;
+	const data: ListResponse | undefined = await response.json();
+	if (!data || !Array.isArray(data.models)) {
+		throw new Error('Failed to parse Ollama tags', { cause: data });
+	}
+
+	// Sort alphabetically
+	data.models = data.models.sort((a, b) => {
+		const nameA = a.name;
+		const nameB = b.name;
```
fmaclen, line 20: There as a mistake in `type OllamaModel` where `model: string` didn't exist, so I updated the compare function to `a.name`/`b.name`.

We are now using the type definitions directly from `ollama-js`, in this case `ListResponse`.

fmaclen:
@binarynoise thank you for the PR!
Hoping to get #124 out soon.

PR #130, opened 2024-07-31
feat: add Desktop release for macOS, Windows & Linux

### LINKED ISSUE #64: Add desktop release
- I'm leaning towards Electron mostly to keep the codebase in the same language.
- Do we want to build the app as a static site and run that inside the window? Or do we want to host a local Hollama server?
- Code-signing: 
  - https://www.electron.build/code-signing
  - https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/
  - https://help.apple.com/xcode/mac/current/#/dev154b28f09

GregoMac1:
Closes #64 

image
image


electron/main.js
```diff
@@ -0,0 +1,60 @@
...
+		console.log('Server started on port', PORT);
+		done();
+	});
+};
+
+function createWindow() {
+	// Create the browser window.
+	const mainWindow = new BrowserWindow({
+		width: 800,
+		height: 600,
```
fmaclen, line 25:
```suggestion
		width: 1280,
		height: 800,
		minWidth: 400,
		minHeight: 640
```

electron/main.js
```diff
@@ -0,0 +1,60 @@
...
+		webPreferences: {
+			preload: path.join(electron, 'preload.js')
+		}
```
fmaclen, line 28: I think we can skip the preload script

package.json
```diff
@@ -25,11 +29,15 @@
 		"@typescript-eslint/parser": "^7.0.0",
 		"autoprefixer": "^10.4.16",
 		"codemirror": "^6.0.1",
+		"cors": "^2.8.5",
```
fmaclen, line 32: Once we remove `express`, we no longer need `cors`.

electron/main.js
```diff
@@ -0,0 +1,60 @@
...
+	});
+
+	// and load the index.html of the app.
+	//   mainWindow.loadFile('index.html')
```
fmaclen, line 32: Let's remove the boiler plate comments.

svelte.config.js
```diff
@@ -3,15 +3,28 @@ dotenv.config();
...
+} else {
+	const adapterConfig = {
+		// See below for an explanation of these options
+		routes: {
+			include: ['/*'],
+			exclude: ['<all>']
+		}
+	};
+
+	if (process.env.ADAPTER === 'docker-node') adapter = adapterNode(adapterConfig);
```
fmaclen, line 25: > We have 2 build pipelines: `adapter-node` and `adapter-cloudflare`, we use [`adapter-node` for the Docker release](https://github.com/fmaclen/hollama/blob/main/svelte.config.js#L27-L29) and we would also use that one for Electron.

```typescript
if (['docker-node', 'electron-node'].includes(process.env.ADAPTER)) adapter = adapterNode(adapterConfig);
```

fmaclen (changes requested):
Haven't had a chance to actually try the build yet but there are a few things we need to fix first.

I'll provide the remaining assets (icons, "loading.html", etc) later.

svelte.config.js
GregoMac1: About this, I couldn't make Electron work using `adapter-node`, that's why I used `adapter-static`. I can give it another try.

fmaclen: Canutin desktop [works with `adapter-node`](https://github.com/fmaclen/canutin-desktop/blob/master/sveltekit/svelte.config.js) so this one should too; but it's possible I'm missing some other required config.

Did you run `ADAPTER="electron-node" npm run build`?
Maybe the build you tried used was made with `adapter-cloudflare` (which is the default adapter)?

Make the change, let me know what errors you are getting and I'll give you a hand.

.github/workflows/publish.yml
```diff
@@ -55,6 +55,33 @@ jobs:
...
+        run: npm ci
+
+      - name: Build & release Electron app
+        uses: coparse-inc/action-electron-builder@v1.0.0
+        with:
+          github_token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+          build_script_name: 'electron:build'
+          release: true
+        env:
+          ADAPTER: electron-node
```
fmaclen, line 83: I updated `build_script_name` so it runs the script you created:
```json
"electron:build": "ADAPTER=electron-node vite build && electron-builder"
```

However, I'm also intentionally adding the `ADAPTER` environment variable in this file because `window-latest` doesn't allow you to add environment variables as part of a command.

electron/main.js
```diff
@@ -0,0 +1,40 @@
+import { join } from "path";
```
fmaclen, line 1: I cleaned up this file a bit:
- Removed most of the boilerplate comments
- Removed debugging `console.log`'s, the ones that remain were intentionally left and should only be visible in `dev` mode.

electron/main.js
```diff
@@ -0,0 +1,40 @@
...
+import { app, BrowserWindow, utilityProcess } from 'electron';
+
+const isAppPackaged = app.getVersion() !== '0.0.0-dev';
+const hollamaHost = isAppPackaged ? '0.0.0.0' : '127.0.0.1';
+const hollamaPort = isAppPackaged ? '4173' : '5173';
```
fmaclen, line 6: I don't love this implementation but basically we assume we are in a `dev` environment if the version of `package.json` is `0.0.0-dev`, which is the default value.

This version get's overwritten by [`update-package-json.cjs`](https://github.com/fmaclen/hollama/blob/main/.github/workflows/publish/update-package-json.cjs) during the `publish.yml` GH action.

In `dev` mode we want to intentionally run Hollama the way we normally run it: `npm run dev`, which listens on `5173`.

When the _app is packaged_ (`production`), we bind the server to `0.0.0.0` so it can be accessed from other devices in the same LAN, and we use `4173` which is Vite's default "build" port.

Eventually we would probably need to allow the user to configure all of this stuff but it's fine for now.

electron/main.js
```diff
@@ -0,0 +1,40 @@
...
+		minWidth: 400,
+		minHeight: 640
+	});
+
+	mainWindow.loadURL(`http://${hollamaHost}:${hollamaPort}`);
+}
+
+app.whenReady().then(() => {
+	if (isAppPackaged) {
+		const utility = utilityProcess.fork(join(app.getAppPath(), "build", "index.js"), {
```
fmaclen, line 21: I simplified the path syntax for finding the `index.js` that starts the server.

package.json
fmaclen: These removals were marked as resolved but were never deleted.

package.json
```diff
@@ -52,5 +58,27 @@
...
+		},
+		"win": {
+			"target": "nsis",
+			"icon": "electron/resources/icons/win/icon.ico"
+		},
+		"linux": {
+			"target": "AppImage",
+			"icon": "electron/resources/icons/png/1024x1024.png"
+		}
+	}
```
fmaclen, line 83: Didn't really make changes to your config other than specifying the paths to the icons.

fmaclen (changes requested):
Ended up making some of the changes myself, I spent several hours trying to simplify the workflow and it was a lot easier to just commit my changes directly, so I left you some notes on what changes I made and why.

So far, I only tested the _dev_ and _packaged_ builds on macOS and it works fine.

Next, I will do some QA on the Windows & Linux releases.

After that, I will QA the Github Action in a separate repo to make sure it's all running smoothly.

electron/main.js
```diff
@@ -0,0 +1,43 @@
...
+		} else {
+			console.warn('##### Running Electron in development mode');
+			console.log('##### Run `npm run dev` to start the Hollama server separately');
+			createWindow();
+		}
+
+		// macOS: Open a window if none are open
+		app.on('activate', function () {
+			if (BrowserWindow.getAllWindows().length === 0) createWindow();
+		});
```
fmaclen, line 36: `utility.on('message', ... )` never get's called because the SvelteKit server never sends such signal.
The only reason why this was working on macOS is because of the behavior in `app.on('activate', ... )`.

electron/main.js
```diff
@@ -0,0 +1,43 @@
...
+
+	mainWindow.loadURL(`http://${hollamaHost}:${hollamaPort}`);
+}
+
+app
+	.whenReady()
+	.then(() => {
+		if (app.isPackaged) {
+			const utility = utilityProcess.fork(join(app.getAppPath(), 'build', 'index.js'), {
+				env: { ...process.env, HOST: hollamaHost, PORT: hollamaPort }
```
fmaclen, line 23: Need to spread `...process.env` or the server fails to initialize in Windows.

.github/workflows/publish.yml
```diff
@@ -55,6 +55,33 @@ jobs:
           accountId: ${{ secrets.CF_ACCOUNT_ID }}
           command: pages publish .svelte-kit/cloudflare --project-name hollama
 
+  electron:
```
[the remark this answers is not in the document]
fmaclen: The job runs fine 👍

electron/main.js
```diff
@@ -0,0 +1,89 @@
...
+
+			socket.connect(port, HOLLAMA_HOST, () => {
+				socket.destroy();
+				resolve();
+			});
+		}
+
+		tryConnection();
+	});
+}
```
fmaclen, line 54: We use this function to check if the server is ready once per second, up-to 10 times.

PR #145, opened 2024-08-06
fix: abort completion if user confirms to change the current session

GregoMac1:
* fix manually aborting a session, it was not working
* if the user clicks on another session while a completion is being written, they have to confirm
* if the user confirms, the completion is aborted before changing the session


src/routes/sessions/[id]/+page.svelte
```diff
@@ -202,6 +204,17 @@
 		session.messages = [...session.messages, message];
 	}
 
+	beforeNavigate((navigation) => {
+		if (completion) {
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:208]: Add a guard clause:`if (completion) return;`

src/routes/sessions/[id]/+page.svelte
```diff
@@ -202,6 +204,17 @@
...
+			const userConfirmed = confirm('A completion is in progress. Do you really want to leave?');
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:209]: `Are you sure you want to leave?\nThe completion in progress will stop`

src/routes/sessions/[id]/+page.svelte
```diff
@@ -34,6 +34,7 @@
 	import ButtonDelete from '$lib/components/ButtonDelete.svelte';
 	import Metadata from '$lib/components/Metadata.svelte';
 	import Head from '$lib/components/Head.svelte';
+	import { beforeNavigate } from '$app/navigation';
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:37]: I usually try to keep the 3rd-party library imports at the top of the file (such as this one from Svelte).
I was hoping prettier would sort them automatically.

fmaclen (changes requested):
Looks great 👍 
Couple of minor copy changes and syntax cleanup.

src/routes/sessions/[id]/+page.svelte
Comment (fmaclen) [reply src/routes/sessions/[id]/+page.svelte:208]: In general I try to add guard clauses in cases where it would reduce the indentation of whatever is inside the `if { .. }` statements.

fmaclen:
Encontré un bug, cuando paras un `completion` en curso y haces click en algun otro link te pregunta si estas seguro de abandonar:

https://github.com/user-attachments/assets/a452738f-2dfa-49a8-8987-6040b3ee3c38

GregoMac1:
@fmaclen the code is ready!

About the test, I initially didn't know how to emulate the behavior of the 'pending' requests, so I asked Claude for help and I ended up with [this](https://github.com/fmaclen/hollama/pull/145/files#diff-b728858cc9ac6c41fb9cebc11f95efebff4b3708eefe0b569f765fee9b52998c) test case.

As far as I understand, it should work, but it doesn't. It fails because, at line 609, it expects the completion to be started, but it is not. I think it has to be something about the stream response emulated at line 591.

Do you see something wrong?

src/routes/sessions/[id]/+page.svelte
```diff
@@ -202,6 +200,41 @@
...
+		prompt = promptCached;
+		promptCached = '';
+	}
+
+	function abortOllama() {
+		ollamaInstance?.abort();
+		completion = '';
+		// Remove the "incomplete" AI response
+		session.messages = session.messages.slice(0, -1);
+	}
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:214]: A few things we can improve about the stop response logic:
- Let's rename this function to `stopResponse()` so it's more descriptive of the _user behavior_, rather than the _technical implementation_.
- There is some code repetition with the `isWaitingForResponse` logic, so let's move it inside `stopResponse()`
- Once we move `if (isWaitingForResponse)`, we will also need to pass an optional argument to handle the beforeNavigate logic: `stopResponse(navigation?: BeforeNavigate)` for `navigate.cancel()`
- Because the toast alert is tied to the abort behavior we should move the toast event to this function
  - Also, it should be a `toast.warn()` because it's not an error.
- Should we also reset the prompt if you navigate out? if so, we might want to call `resetPrompt()` directly inside `stopResponse()`

Should look more or less like this:

```typescript
	import type { BeforeNavigate } from '@sveltejs/kit';

	function stopResponse(navigation?: BeforeNavigate) {
		if (isWaitingForResponse) {
			toast.warning("Can't stop completion until the first response from Ollama is received");
			navigation?.cancel();
			return;
		}

		resetPrompt();
		ollamaInstance?.abort();
		completion = '';
		session.messages = session.messages.slice(0, -1); // Remove the "incomplete" AI response
	}
```

src/routes/sessions/[id]/+page.svelte
```diff
@@ -202,6 +200,41 @@
...
+			const userConfirmed = confirm(
+				'Are you sure you want to leave?\nThe completion in progress will stop'
+			);
+			if (userConfirmed) {
+				abortOllama();
+				return;
+			}
+			navigation.cancel();
+		}
+	});
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:236]: And we can simplify this function as:

```suggestion
	beforeNavigate((navigation) => {
		if (!completion) return;
		const userConfirmed = confirm(
			'Are you sure you want to leave?\nThe completion in progress will stop'
		);
		if (!userConfirmed) navigation.cancel();
		stopResponse(navigation);
	});
```

src/routes/sessions/[id]/+page.svelte
```diff
@@ -202,6 +200,41 @@
...
+	function abortOllama() {
+		ollamaInstance?.abort();
+		completion = '';
+		// Remove the "incomplete" AI response
+		session.messages = session.messages.slice(0, -1);
+	}
+
+	function showAbortError() {
+		toast.error("Can't stop completion until the first response from Ollama is received");
+	}
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:218]: This can be deleted

src/routes/sessions/[id]/+page.svelte
```diff
@@ -318,8 +351,13 @@
...
 									variant="outline"
 									on:click={() => {
-										abortController.abort();
+										if (isWaitingForResponse) {
+											showAbortError();
+											return;
+										}
+										abortOllama();
 										resetPrompt();
+										abortController.abort();
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:360]: We need to remove all instances of the old `abortController`.

src/routes/sessions/[id]/+page.svelte
```diff
@@ -318,8 +351,13 @@
 									title="Stop response"
 									variant="outline"
 									on:click={() => {
-										abortController.abort();
+										if (isWaitingForResponse) {
+											showAbortError();
+											return;
+										}
+										abortOllama();
 										resetPrompt();
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:359]: All of this logic can now be `on:click={() => stopResponse()}`

tests/session.test.ts
```diff
@@ -546,6 +546,86 @@ test.describe('Session', () => {
...
+			// Submit prompt
+			await promptTextarea.fill('Test prompt');
+			await page.getByText('Run').click();
+		}
+
+		// Mock a slow response
+		await page.route('**/api/chat', async (route) => {
+			await new Promise((resolve) => setTimeout(resolve, 1000)); // Simulate pending state
+			await route.fulfill({ status: 200, body: '{"message": {"content": "Loading..."}}' });
+		});
```
fmaclen, line 572: > As far as I understand, it should work, but it doesn't. It fails because, at line 609, it expects the completion to be started, but it is not. I think it has to be something about the stream response emulated at line 591.

When I run the test I get an error on line 587:
```
await expect(errorMessage).toBeVisible();
```

I think the issue is that the first `// Mock a slow response` is not a streamed response.
If you use the same mocked streamed response as `// Mock completion start` on line 591 it successfully asserts the warning.

tests/session.test.ts
```diff
@@ -546,6 +546,86 @@ test.describe('Session', () => {
...
 	});
 
+	test('handles abort and navigation during completion', async ({ page }) => {
+		const aiMessage = page.locator('article', { hasText: 'Assistant' });
+		const errorMessage = page.getByText(
+			"Can't stop completion until the first response from Ollama is received"
+		);
+
+		async function setupAndRunPrompt() {
+			// Setup
```
fmaclen, line 556: > As far as I understand, it should work, but it doesn't. It fails because, at line 609, it expects the completion to be started, but it is not. I think it has to be something about the stream response emulated at line 591.

I see some issues with the test but before I dig deeper we must fix the implementation because it will likely affect the behavior of the test.

That being said, we should probably do some cleanup here as well:
- `errorMessage` > `warningMessage`
- `"Can't stop completion until the first response from Ollama is received, please wait..."`

The fact that we need a `setupAndRunPrompt()` function feels like a code smell. A better way of handling this is to split the test into multiple smaller tests that are isolated from one-another, for example we could do:

```typescript
	test.describe('aborting completions', () => {
		test.beforeEach(async ({ page }) => {
			// move logic from setupAndRunPrompt() here
		});

		test("can't stop a while waiting for response", async ({ page }) => { ... });
		test('can stop a response in progress', async ({ page }) => { ... });
		test("can't navigate away from session while waiting for response", async ({ page }) => { ... });
		test('can navigate away from session while a response is in progress', async ({ page }) => { ... });
	});
```

tests/session.test.ts
```diff
@@ -546,6 +546,86 @@ test.describe('Session', () => {
...
+		// Mock completion start
+		await page.route('**/api/chat', async (route) => {
+			const encoder = new TextEncoder();
+			const readable = new ReadableStream({
+				start(controller) {
+					controller.enqueue(encoder.encode('{"message": {"content": "Starting response..."}}\n'));
+					// Keep the stream open
+				}
+			});
+			// @ts-ignore
```
fmaclen, line 599: Can we fix the typing error without using annotations/comments?

fmaclen (changes requested):
The behavior appears to work as expected but there is some cleanup we need to do.
Make sure to remove old logic that no longer applies.

tests/session.test.ts
fmaclen: Also, whenever you run into weird issues with the tests you can install the [Playwright extension for VSCode](https://playwright.dev/docs/getting-started-vscode) and run it in debug mode: https://playwright.dev/docs/debug#run-in-debug-mode

GregoMac1:
I've changed the code and I've split the tests but I still could not make them work. I'll dig deeper tomorrow.

src/routes/sessions/[id]/+page.svelte
```diff
@@ -202,6 +200,39 @@
...
+	beforeNavigate((navigation) => {
+		if (!completion) return;
+		const userConfirmed = confirm(
+			'Are you sure you want to leave?\nThe completion in progress will stop'
+		);
+		if (!userConfirmed) {
+			navigation.cancel();
+			return;
+		}
+		stopResponse(navigation);
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:233]: Can we revert this syntax so it's less brittle?
```suggestion
		if (userConfirmed) {
			stopResponse(navigation);
			return;
		}
		navigation.cancel();
```

fmaclen:
> I've split the tests but I still could not make them work

There's a few issues with the tests:
- The shape of the streamed mocked responses should match the [Ollama API](https://github.com/fmaclen/hollama/pull/145#discussion_r1717031813)
- But more importantly, Playwright doesn't appear to support mocking streamed responses.
  - There is [an issue with a suggested workaround](https://github.com/microsoft/playwright/issues/15353) but haven't tried it to see if it works.
  - Here's another potential solution from StackOverflow: https://stackoverflow.com/questions/77030676/how-to-stub-mock-a-streamed-response-from-chatgpt
- The test `can navigate away from session while a response is in progress` looks incomplete to me, somewhere there we should be "accepting" the confirmation dialog.
- The test `can start a new session, choose a model and stop a completion in progress` is now failing, we can probably move the `stop response` UI assertions from that test into the new ones.
- Pull the latest changes from `main` to fix the test `seed data and take screenshots for README.md `.

GregoMac1:
Thanks for the feedback!

> * The shape of the streamed mocked responses should match the [Ollama API](https://github.com/fmaclen/hollama/pull/145#discussion_r1717031813)

About this, I followed the shape of the real responses from Ollama in the app, which look like this:
image
Instead of `response`, the replies contain `message: { role: string, content: string }`.

I'm still working on the rest of tests!

src/routes/sessions/[id]/+page.svelte
Comment (GregoMac1) [reply src/routes/sessions/[id]/+page.svelte]: I'm afraid this refactored code is not working as it did before. If I click `Run` and then I try to navigate to another page, completion is still `= ''` so the `beforeNavigate` function returns immediately and it does not show the warning toast.

Comment (fmaclen) [reply src/routes/sessions/[id]/+page.svelte]: Understood, that makes sense.

fmaclen:
> Instead of response, the replies contain message: `{ role: string, content: string }`.

Sorry, you are right. I was looking at the `/api/generate` streamed response instead of `/api/chat`. My bad!

tests/session.test.ts
```diff
@@ -634,4 +636,61 @@ test.describe('Session', () => {
...
+		await mockStreamedCompletionResponse(page, MOCK_SESSION_1_RESPONSE_1);
+		await promptTextarea.fill('Who would win in a fight between Emma Watson and Jessica Alba?');
+		await runButton.click();
+
+		// Wait for the completion to start
+		await expect(page.getByText('...')).toBeVisible();
+
+		// Set up dialog handler to cancel
+		dialogHandler = async (dialog) => {
+			expect(dialog.message()).toContain('Are you sure you want to leave?');
```
fmaclen, line 661: Let's test the entire string:
```
'Are you sure you want to leave?\nThe completion in progress will stop'
```

tests/session.test.ts
```diff
@@ -634,4 +636,61 @@ test.describe('Session', () => {
...
+		let dialogHandler: (dialog: Dialog) => Promise<void>;
+		page.on('dialog', async (dialog) => await dialogHandler(dialog));
+
+		await page.goto('/');
+		await chooseModelFromSettings(page, MOCK_API_TAGS_RESPONSE.models[0].name);
+		await page.getByText('Sessions', { exact: true }).click();
+		await page.getByTestId('new-session').click();
+
+		// Start a streamed completion
+		await mockStreamedCompletionResponse(page, MOCK_SESSION_1_RESPONSE_1);
```
fmaclen, line 652: Try replacing this with:

```typescript
await page.route('**/chat', () => {});
```

fmaclen:
@GregoMac1 I updated this PR with the latest merges on `main` and fixed the conflicts:
https://github.com/fmaclen/hollama/pull/156
https://github.com/fmaclen/hollama/pull/157

fmaclen (approved):
LGTM 🚀

PR #153, opened 2024-08-12
chore: add localization to UI and components

### LINKED ISSUE #150: Localize labels in UI components
I'ved used `typesafe-i18n` in the [past](https://github.com/fmaclen/julia-sanfrancisco/tree/main/src/i18n), but we might as well use [`i18next`](https://github.com/i18next/i18next?tab=readme-ov-file), how bad [could it be](https://github.com/open-webui/open-webui/blob/main/package.json#L65C4-L65C11)?

### LINKED ISSUE #151: Translate localized component labels to Spanish
~~Blocked by #150 #171~~ 🎉 

- [x] Translate strings to es
- [x] Detect language from the browser
- [x] Switch the language if it's available, otherwise fallback to **English**
- [x] Add a select field in **Settings** for manually switching the language
- [x] Save the language choice (if made) to the `localStorage`

GregoMac1:
Closes #150


fmaclen:
Couple of edits:
- Updated PR title to `chore:` prefix so it doesn't trigger a release
- I removed `Closes #151`, we should do that in a separate PR

GregoMac1:
@fmaclen I'm still working on this, I still haven't tried all the possible cases to see is every text is showing correclty.

Also, I've just merged `main` and I can see that there are new texts that I have to localize.

However, I'd like to ask you to please review it to see if the current approach is correct or if there's anything that needs to be changed.

PS: About the failing test, I don't exactly know why it fails. When I run `npm run test` on my computer, all tests finish successfully. Could it be something about the screenshots' size?

src/lib/i18n.ts
```diff
@@ -0,0 +1,86 @@
+import i18next from 'i18next';
+import { createI18nStore } from 'svelte-i18next';
+
+i18next.init({
+	lng: 'en',
+	resources: {
+		en: {
+			translation: {
+				nav: {
```
fmaclen, line 9: Let's rename the `nav` key to `sidebar`

src/lib/i18n.ts
```diff
@@ -0,0 +1,86 @@
...
+					knowledge: 'Knowledge',
+					settings: 'Settings',
+					motd: 'Motd',
+					theme: { dark: 'Dark', light: 'Light' }
+				},
+				sessions: {
+					pageTitle: 'Sessions',
+					new: 'New session',
+					empty: 'No sessions',
+					session: 'Session',
```
fmaclen, line 20: Let's get rid of all the `pageTitle`'s.

Instead of `pageTitle: Sessions` and `session: Session` let's use the [pluralization syntax](https://www.i18next.com/translation-function/plurals):

```
one: Session
other: Sessions
```

```
<Head title={$i18n.t('sessions', count: 0)} />
```

src/lib/i18n.ts
```diff
@@ -0,0 +1,86 @@
...
+					connected: 'connected',
+					dangerZone: 'Danger zone',
+					deleteAllSessions: 'Delete all sessions',
+					deleteAllKnowledge: 'Delete all knowledge',
+					deleteServerSettings: 'Delete server settings',
+					version: 'Version'
+				},
+				motd: {
+					pageTitle: 'Message of the day'
+				},
```
fmaclen, line 63: Since this group only has one value, let's just do: `motd: 'Message of the day'`

src/lib/i18n.ts
```diff
@@ -0,0 +1,86 @@
...
+					noSessionChosen: 'Create a new session or choose an existing one from the list',
+					promptPlaceholder: 'Write literally anything',
+					cantConnectToOllama: "Can't connect to Ollama server",
+					stopResponse: 'Stop response',
+					systemPrompt: 'System prompt',
+					goToKnowledge: 'Go to knowledge',
+					writePromptToStart: 'Write a prompt to start a new session',
+					ollamaConnectionError: "Couldn't connect to Ollama. Is the [server running](/settings)?",
+					genericError: 'Sorry, something went wrong.\n```\n${{error}}\n```',
+					somethingWentWrong: 'Sorry, something went wrong.'
```
fmaclen, line 35: Let's move all of the errors: `cantConnectToOllama`, `ollamaConnectionError`, etc... to `errors: { }`

src/lib/i18n.ts
```diff
@@ -0,0 +1,86 @@
...
+				},
+				common: {
+					copy: 'Copy',
+					retry: 'Retry',
+					confirmDeletion: 'Confirm deletion',
+					dismiss: 'Dismiss',
+					you: 'You',
+					assistant: 'Assistant',
+					system: 'System'
+				}
```
fmaclen, line 76: Let's move all of these to be direct children of `en: { translation: { }  }`

fmaclen (changes requested):
I'm not sure if you are still working on this but I left you some note changes.

In my experience with large projects that have a lot of localized strings grouping keys by "section" doesn't work very as the project grows.

Mainly because you end up with generic keys such as `name`, `save`, `delete`, `server`, etc that need to be used in other "sections", which if you are not super diligent about keeping things organized you quickly run into duplicates.

I think your grouping is okay for now, but we will probably start moving most of those keys up to the parent `en` tree.

fmaclen:
> I was wondering on how to translate [this](https://github.com/fmaclen/hollama/blob/e80847d314caf5293c96818e3b60c42a5d40edf4/src/routes/settings/%2Bpage.svelte#L148-L179)

Yeah, these cases are tricky because we would need to interpolate components in the middle of a string.

I think the realistic solution to this issue is to re-write the sentences in such a way so the labels in the components can stand alone.

For example:

```
<p class="p">
  {{ Change your server settings to allow connections: }}
  <Badge capitalize={false}>OLLAMA_ORIGINS={ollamaURL.origin}</Badge>
  <a>{{ See docs }}</a>.
  {{ Also check no browser extensions are blocking the connection. }}
</p>
```

It sounds a bit more "robotic" but I think it's fine for now.

src/lib/i18n.ts
```diff
@@ -0,0 +1,82 @@
...
+i18next.init({
+	lng: 'en',
+	resources: {
+		en: {
+			translation: {
+				session_one: 'Session',
+				session_other: 'Sessions',
+				knowledge_one: 'Knowledge',
+				knowledge_other: 'Knowledge',
+				settings: 'Settings',
```
fmaclen, line 13: You don't need `sessions` and `sessionPage`, you can just do:
```
sessions: {
  one: "Session",
  other: "Sessions"
}
```

```
settings: {
  one: "Setting",
}
```

fmaclen (changes requested):
- Minor change in the `sectionPage` naming convention.
- We are missing the localized sentences in the **Settings** page, I left you some suggestions in [this comment](https://github.com/fmaclen/hollama/pull/153#issuecomment-2307422861).
- We are also missing some dialog confirmations, errors, toast messages

src/lib/i18n.ts
fmaclen: Hang on, this is how I usually pluralize things in other projects but that may be different in this library.

src/lib/i18n.ts
```diff
@@ -0,0 +1,82 @@
...
+				motd: 'Message of the day',
+				sidebar: {
+					sessions: 'Sessions',
+					knowledge: 'Knowledge',
+					settings: 'Settings',
+					motd: 'Motd',
```
fmaclen, line 19: Let's not repeat keys with the same value, grab them from the `sessions` and `knowledge` keys.

src/lib/i18n.ts
fmaclen: I'll likely end up moving most of the keys as direct children of `translation: { }`

src/routes/+layout.svelte
```diff
@@ -63,17 +64,17 @@
 				{:else if href === '/motd'}
 					<NotebookText class="h-4 w-4" />
 				{/if}
-				{text}
+				{$i18n.t(`sidebar.${text.toLowerCase()}`)}
```
fmaclen, line 67: Here I don't think you need to call `text.toLowerCase()`, simply update the values in `const SITEMAP` so they are lowercase by default.

GregoMac1:
@fmaclen I've translated the remaining texts.

> And in this example we can keep the `<code>` tags in the translated strings:
> 
> ```
> <p>
>   {{ If you want to connect to an Ollama server that is not available on <code class="code">localhost</code> or <code class="code">127.0.0.1</code> you can try: }}
>   <a>{{ Creating a tunnel }}</a>
>   <a>{{ Allowing mixed content }}</a>
> </p>
> ```
About this quote of [this comment](https://github.com/fmaclen/hollama/pull/153#issuecomment-2307422861), I've tried it in several ways using the [escaping options](https://www.i18next.com/translation-function/interpolation#unescape), but I couldn't get to the desired result (image below), so for now I left the unformatted text in the translation.

image


As for the rest of texts with interpolated `html` tags, please let me know if you'd wish something to be changed.

fmaclen:
> I've tried it in several ways using the [escaping options](https://www.i18next.com/translation-function/interpolation#unescape), but I couldn't get to the desired result

Try the svelte syntax [`{@html}`](https://svelte.dev/docs/special-tags#html), [here's an example](https://github.com/fmaclen/hollama/blob/main/src/lib/components/Markdown.svelte#L49).

fmaclen (approved):
I think everything else looks good, once you fix the escaping issue let's get this one merged.

GregoMac1:
Thank you! It worked!

> Try the svelte syntax [`{@html}`](https://svelte.dev/docs/special-tags#html), [here's an example](https://github.com/fmaclen/hollama/blob/main/src/lib/components/Markdown.svelte#L49).

About the example, I think you can get rid of the 'hack' by using `eslint-disable-next-line` instead of `eslint-disable-line` (check [this](https://github.com/fmaclen/hollama/pull/153/commits/d3b0afce3aec19c27d58b56165b488cf99a9c2bc) commit), although I'm not sure if that was exactly the problem.

PR #154, opened 2024-08-14
feat: add searchable select (svelte-select)

GregoMac1:
Allow users to search and filter between the options of a Select component.


fmaclen:
Decided to go with #181

Issue #126, opened 2024-07-28
Output from multiple sessions gets jumbled together


Infinoid:
If you have two session tabs where the AI is responding to two prompts at the same time, their outputs get jumbled together.

I think Hollama has two HTTP connections open, receiving chunked-encoding responses from each, but those responses are both added to whichever session tab is currently selected (active) in the UI, not the session tab that the prompt came from.

To reproduce:

1. Open a new session.  Ask a short question, like "Does this chat work?", to create the session tab in the session list.  Wait for the answer to complete.
2. Open a second session in the same way.  Wait for the answer to complete.
3. Type a longer question in the first session tab.
4. While it is being answered, type a longer question in the second session tab.

For example, I had two `llama3.1:8b` sessions open, and asked each of them to generate a short story at the same time, one about a grasshopper and one about a carpenter.  As the responses were generated, I saw jumbled output in the second session tab that looks like this:

> adow filled with vibrant wildflowersHere and's buzzing a bees short. story His for days you were:
> 
> spent---
> 
> leKapingait fromo blade Yam toato blade was, just sn anatching average tasty guy insects from for modern lunch-day, Tokyo and, b workingasking as in a the carp warmenter sunlight in.

After the two stories finished generating:
1. The first session tab has no AI text boxes in it, just my prompt with no response.
2. The second session tab had two AI text boxes in it, which seemed to contain two copies of the same jumbled text.

I suspect that if I had selected the first session tab, the two text boxes would have ended up in that tab, instead.

Infinoid:
Here are some other things I checked:
* It doesn't seem to matter which model is used; their outputs are all jumbled in the same way.
* I am able to send two prompts to my `ollama` service at the same time using `curl`, and the responses look just fine.
* In Chrome Devtools, I can see two `/api/generate` requests running in parallel, and their responses both look fine.

In both curl and chrome devtools, if I ignore all of the JSON and just look at the "response" fields, each story is readable.  So I think it's a problem with the UI applying responses to the wrong session tab.

fmaclen:
@Infinoid thanks for the bug report.

I'm currently [working on a PR](https://github.com/fmaclen/hollama/pull/125) to fix another issue that I think would also address this one (though maybe partially).

In fact, can you try it out?
https://api-chat-endpoint.hollama.pages.dev/

The implementation in that environment is not 100% finished but I just did a quick test with 2 completions running simultaneously in 2 tabs and they didn't get jumbled 🤞

[...]

Issue #103, opened 2024-07-20
Session breaks upon parsing context


binarynoise:
> Sorry, something went wrong.
> ```
> SyntaxError: JSON.parse: end of data when ',' or ']' was expected at line 1 column x of the JSON data
> ```

This is independent from the knowledge as system prompt experiment.

Here is the network traffic for that session as [har file](http://www.softwareishard.com/blog/har-12-spec/): [localhost_Archive [24-07-20 17-45-41].har.json](https://github.com/user-attachments/files/16321571/localhost_Archive.24-07-20.17-45-41.har.json). You can load it into Firefox or Chrome to see the details.

As far as I can tell, ollama does send a valid response (in contrast to #97) but hollama for some reason fails to parse the last line with the context. The x varies wildly ~~around 30k~~.
It happens around the 5th or 6th response for me with different models and prompts.

As a consequence, the conversation's context is lost and a new conversation begins on retrying the last message. Sometimes not everything gets lost but the conversation is broken anyway.

As a workaround, would it be possible to hold onto the last valid context and use that when retrying the message?

Switching to another conversation and coming back to retry does actually trigger #97.

binarynoise:
Or just save the returned context for each message to be able to start over at a given point when the model starts generating nonsense

fmaclen:
Thanks, I was able to replicate the issue.

I'm not entirely sure yet but I suspect is because we are trying to format the completion in "small chunks" as Ollama streams them, one of the chunks is causing a parser to break.

> As a workaround, would it be possible to hold onto the last valid context and use that when retrying the message?

Yeah, I wanted to implement something like this so messages can be retried or even edited. #9

fmaclen:
Here's what I think it's happening, when we catch an error during a completion (any kind of error) the prompt is reset so the user can try again, but when it re-submits the prompt the context array with all of the tokens is lost.

fmaclen:
I don't think we can predict when the JSON parsing will fail, but whenever there is any kind of error the UI should hint a clear path on how to fix it (if possible).

In a large number of cases I expect simply "retrying" should fix most issues: https://github.com/fmaclen/hollama/issues/9#issuecomment-2241328191

binarynoise:
#9 doesn't help at all for this problem. Regenerating the failing message or even one earlier will cause the same error again.

fmaclen:
@binarynoise true, #103 won't be fixed by #106.

[...]

PR #156, opened 2024-08-17
fix: revert `ollama-js` implementation with `fetch`

fmaclen:
- Replaces `ollama-js` with `fetch()` using the old implementation for `abortController`
- We are keeping `ollama-js` as a source for type definitinos
- Replaces custom type definitions `OllamaModel`, `OllamaTagResponse` with `import type { ListResponse } from 'ollama/browser';`


src/lib/ollama.ts
```diff
@@ -1,25 +1,18 @@
...
-		quantization_level: string;
-	};
-};
+export async function ollamaChat(payload: ChatRequest, abortSignal: AbortSignal) {
+	const settings = get(settingsStore);
+	if (!settings) throw new Error('No Ollama server specified');
 
-export type OllamaTagResponse = {
-	models: OllamaModel[];
-};
```
fmaclen, line 22: Minor cleanup, instead of having our own custom types we can simply use the types as defined by `ollama-js`. One less thing to maintain.

src/routes/sessions/[id]/+page.svelte
```diff
@@ -44,14 +43,13 @@
 	let knowledge: Knowledge | null;
 	let session: Session;
 	let completion: string;
+	let abortController: AbortController;
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:46]: Restored the old `abortController`

src/routes/sessions/[id]/+page.svelte
```diff
@@ -207,17 +212,9 @@
...
 
-	function stopResponse(navigation?: BeforeNavigate) {
-		if (isWaitingForResponse) {
-			toast.warning(
-				"Can't stop completion until the first response from Ollama is received, please wait..."
-			);
-			navigation?.cancel();
-			return;
-		}
-
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:218]: We can now stop the request at any time so we no longer need this _hacky_ warning.

src/lib/ollama.ts
```diff
@@ -1,25 +1,46 @@
...
-
-export type OllamaTagResponse = {
-	models: OllamaModel[];
-};
+export async function ollamaChat(
+  payload: ChatRequest, 
+  abortSignal: AbortSignal, 
+  onChunk: (content: string) => void
+) {
```
fmaclen, line 9: The implementation of `ollamaChat()` is very similar to the old [`ollamaGenerate()`](https://github.com/fmaclen/hollama/pull/125/files#diff-31ae89989234ded3c6f93e8e5fe7cc2c30e93536cc027383faae05989bf79205L45-L55)

src/routes/settings/+page.svelte
```diff
@@ -1,15 +1,16 @@
...
 
 	import Badge from '$lib/components/Badge.svelte';
 	import Button from '$lib/components/Button.svelte';
 	import FieldSelectModel from '$lib/components/FieldSelectModel.svelte';
 	import Fieldset from '$lib/components/Fieldset.svelte';
 	import FieldInput from '$lib/components/FieldInput.svelte';
-
-	import { ollamaTags, type OllamaTagResponse } from '$lib/ollama';
-	import { LOCAL_STORAGE_PREFIX, settingsStore, StorageKey } from '$lib/store';
 	import Head from '$lib/components/Head.svelte';
```
fmaclen, line 14: Re-arranged imports.

tests/session.test.ts
```diff
@@ -525,7 +525,8 @@ test.describe('Session', () => {
 		await expect(page.getByText('Sorry, something went wrong.')).toBeVisible();
 		await expect(
 			page.locator('code', {
-				hasText: 'Error: Did not receive done or success response in stream.'
+				hasText:
+					"SyntaxError: Expected property name or '}' in JSON at position 2 (line 1 column 3)"
 			})
 		).toBeVisible();
```
fmaclen, line 531: We no longer get this error which means that if the JSON syntax breaks we are back to the [old error](https://github.com/fmaclen/hollama/issues/103).

The main issue in #103 was that we were using the `/api/generate` endpoint which broke JSON syntax upon reaching the model's context window, that should no longer be the case 🤞 .

tests/session.test.ts
```diff
@@ -546,94 +547,8 @@ test.describe('Session', () => {
...
-				await route.fulfill({ status: 200, body: readable });
-			});
-
-			await runButton.click();
-
-			await page.getByText('Knowledge', { exact: true }).click();
-			await expect(aiMessage).toContainText('Loading...');
-		});
+	test.skip('can navigate out of session during completion', async () => {
+		// TODO: Add test for navigation during completion
```
fmaclen, line 551: We can get rid of most of the new tests 🥳 

The stop completion functionality is mostly already handled by:
```
can start a new session, choose a model and stop a completion in progress
```

We only need one test that checks that:
- The user confirmation dialog appears while a completion is in progress and the user navigates away
- Cancelling the dialog continues the completion
- Get another completion going, confirm the dialog, check the completion stops and navigates

GregoMac1:
This looks amazing!

I've tested this branch and the only bug that I find is the one we've talked about [here](https://github.com/fmaclen/hollama/pull/145#discussion_r1716925799).

I think we could merge this into #126, fix that problem and then we're ready to go!

GregoMac1:
Now that I think about it again, we should find a way to handle the behaviour of the previous `isWaitingForResponse`. At `beforeNavigate`, we should be able to detect if a request has been sent to `/chat` and completion is still empty. In that case, I think it should throw the alert and the user would have to confirm and abort the response, right?

src/routes/sessions/[id]/+page.svelte
```diff
@@ -410,6 +391,33 @@
...
+	}
+
+	.prompt-editor__stop-icon {
+		@apply opacity-0;
+	}
+
+	.prompt-editor__loading-icon {
+		@apply opacity-100;
 	}
 </style>
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:423]: Adds a loading spinner while completion is in progress:


https://github.com/user-attachments/assets/8658ea7b-cacf-4570-a0b7-c80b05e5d8d6

src/routes/sessions/[id]/+page.svelte
```diff
@@ -44,20 +43,18 @@
...
 	let shouldFocusTextarea = false;
 	let userScrolledUp = false;
-	let ollamaInstance: Ollama | null = null;
-	let isWaitingForResponse = false;
 
 	const shouldConfirmDeletion = writable(false);
 
 	$: session = loadSession(data.id);
 	$: isNewSession = !session?.messages.length;
-	$: isLastMessageFromUser = session?.messages[session.messages.length - 1]?.role === 'user';
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:60]: `isLastMessageFromUser` was only used to determine if there was a completion in progress so we are now using `isCompletionInProgress`

src/routes/sessions/[id]/+page.svelte
```diff
@@ -98,8 +95,6 @@
 		}
 
 		const message: Message = { role: 'user', content: prompt };
-		promptCached = prompt;
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:101]: No need to cache the prompt, we can just grab it from the `session.messages`

GregoMac1 (approved):
Looks great to me!

PR #159, opened 2024-08-19
fix: can manually (or opt-in to automatically) check for updates

fmaclen:
- Can opt-in to check for updates automatically
- If enabled, it checks for updates once a week
- Can manually check for updates regardless of auto-update setting
- It shows update instructions based on the Hollama server environment


fmaclen:
@GregoMac1 just wanted to let you know of some changes this PR introduced that you'll likely interact with:

- `$settingsStore` now has [default values](https://github.com/fmaclen/hollama/pull/159/files#diff-276a0044b7db537e1835eb8b2c20368b8a7437a3fde350198bff4db2b9e418feR61-R74) when it's created, so we no longer need to check if it's `undefined`
- There is a new route [`/api/metadata`](https://github.com/fmaclen/hollama/pull/159/files#diff-cc04b65db56963e89999b3a094103810f1367bd663b437d06049b7c4b4ab11e7R1-R18) endpoint that returns the app's version and the environment
- In `/settings`, I split the different sections into [separate components](https://github.com/fmaclen/hollama/pull/159/files#diff-9756efe8cc16143e4c7b9b60f36591248c9c5c7d46b1d1062be9dce20090f393R12-R14) because having all of that logic together was hard to read.
- I added the new `i18n` keys to the [root of the tree](https://github.com/fmaclen/hollama/pull/159/files#diff-e234bc0b8bc186ee57027897a6aea08d3c733e08fa6eff9d11f7b6a3c4e38446R26-R39).
  - Related: I also created #171 (which we should do before #151)

PR #163, opened 2024-08-25
fix: fix scrolling after submitting a prompt

GregoMac1:
@fmaclen I found the problem. This bug is happening because `handleCompletion` sets `isCompletionInProgress` to `true`, which makes the `...` message visible:

```svelte
{#if isCompletionInProgress}
  <Article message={{ role: 'assistant', content: completion || '...' }} />
{/if}
```

It seems that this render occurs after the `scrollToBottom` invocation. 

I fixed it by using the `requestAnimationFrame` [function](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame), which schedules the scroll action to occur in the next frame after the browser has completed its current rendering tasks.

### DIFF
--- a/src/routes/sessions/[id]/+page.svelte
+++ b/src/routes/sessions/[id]/+page.svelte
@@ -172,9 +172,7 @@
 	async function scrollToBottom(shouldForceScroll = false) {
 		if (!shouldForceScroll && (!messageWindow || userScrolledUp)) return;
 		await tick();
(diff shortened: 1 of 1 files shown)


fmaclen (approved):
> which schedules the scroll action to occur in the next frame after the browser has completed its current rendering tasks

Had no idea that was a thing 😮

PR #169, opened 2024-08-29
fix: general improvements to the UI

### LINKED ISSUE #166: Invert the color scheme of toasts
Text should be white, background should be dark red, currently looks like this:

image
image
image

The idea is that a single shade of color will look good in `light` and `dark` themes.

[`<Toaster richColors={true} position="top-center" />`](https://github.com/fmaclen/hollama/blob/main/src/routes/%2Blayout.svelte#L48)

Here's the playground/docs for `svelte-sonner`, our toast library:
https://svelte-sonner.vercel.app/

### LINKED ISSUE #167: Scrollbars in Windows don't look good in dark mode
image

BukuBuku11:
Closes #167
Closes #166

- [x] styling fixes for **Settings, MOTD, Errors**
- [x] cleaned up styles for `<Fieldset>`


fmaclen:
@BukuBuku11 thanks for opening this PR, I think this is a good start. I never used this library before but it appears to work well.

I tweaked the values for the styling of the scrollbars and abstracted them to a `overflow-scrollbars` class and I updated the `/motd` template with this new class as an example.

The app has multiple instances of `overflow-auto`/`overflow-y-auto` which render these scrollbars, we need to the new helper class to all of them, that way no possible scrollbar goes un-styled.

fmaclen:
Actually, it might make more sense to just extend `overflow-auto`/`overflow-y-auto` so it inherits the scrollbar classes automatically, that way we don't have to remember to add both classes each time we want to use `overflow-auto`.

BukuBuku11:
👍👍

PR #170, opened 2024-08-29
feat: edit user messages

GregoMac1:
Add implementation of the Edit user message feature.

Similar to ChatGPT, when the user edits a message, the old response is replaced with the new one and all newer messages are deleted so that the edited message is the last in the conversation.

For example, in this conversation:

```
User: Prompt 1
AI: Response 1
User: Prompt 2
AI: Response 2
```

If the user edits `Prompt 1`, then `Response 1` is replaced for the new response, and `Prompt 2` and `Response 2` get removed from the session history.


GregoMac1:
@fmaclen please confirm if the described behavior is correct, or if it should be changed. Thanks in advance!

fmaclen:
@GregoMac1 I tried it a bit and it appears to work well, I didn't run into any bugs 👍 

The only thing I noticed is that if you are not paying attention, after clicking the edit button, it's easy to miss that the edited message was copied to the prompt field.

I can think of 2 solutions:
1- When **Edit** is clicked, we switch to the large `<FieldTextEditor>`. This way the change in the UI is so significant that it will be easy to tell what just happened.
2- We add a semi-transparent overlay that obscures the contents of the `<Article>` being edited, with an `<EmptyMessage>` in the middle saying "Edit mode"

I'd try **Option 1** first, I feel like it will be more UX friendly.
Also I think it's easier to implement than **Option 2**, which likely needs a new overlay UI component.

src/lib/i18n.ts
```diff
@@ -32,6 +34,7 @@ i18next.init({
 					content: 'Content',
 					prompt: 'Prompt',
 					run: 'Run',
+					saveAndRun: 'Save & Run',
```
fmaclen, line 37: - Should move the `saveAndRun` key to the root of the tree. Moving forward we want to avoid grouping keys.
- Should be capitalized: `Save & run`

src/routes/sessions/[id]/+page.svelte
```diff
@@ -109,6 +105,29 @@
...
+	async function handleSubmitEditMessage() {
+		if (messageIndexToEdit === null) return;
+
+		session.messages[messageIndexToEdit].content = prompt;
+
+		// Remove all messages after the edited message
+		session.messages = session.messages.slice(0, messageIndexToEdit + 1);
+
+		let payload = { model: session.model, messages: session.messages, stream: true };
+		await handleCompletion(payload);
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:117]: Let's get rid of the single use of `payload` and move the props directly inside the `handleCompletion()`:
```suggestion
		await handleCompletion({ model: session.model, messages: session.messages });
```
Because we used a the `payload` variable we didn't notice that `stream: true` wasn't actually needed.

There are other similar single-uses of `payload` in `handleRetry(payload)` and `handleSubmitNewMessage(payload)`, let's update those as well.

src/routes/sessions/[id]/+page.svelte
```diff
@@ -109,6 +105,29 @@
...
+		prompt = '';
+	}
+
+	$: handleSubmit = () => {
+		if (!prompt) return;
+		isPromptFullscreen = false;
+
+		if (messageIndexToEdit !== null) handleSubmitEditMessage();
+		else handleSubmitNewMessage();
+	};
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:129]: This feels like a code smell, I don't think we need a reactive function to handle this.
We can simplify it by moving the logic to `<ButtonSubmit>`. See note below.

src/routes/sessions/[id]/+page.svelte
```diff
@@ -314,14 +341,27 @@
...
 								hasMetaKey={isPromptFullscreen}
 								disabled={!prompt ||
 									!$settingsStore?.ollamaModels.length ||
 									!$settingsStore?.ollamaModel}
 							>
-								{$i18n.t('sessionsPage.run')}
+								{$i18n.t(
+									messageIndexToEdit !== null ? 'sessionsPage.saveAndRun' : 'sessionsPage.run'
+								)}
 							</ButtonSubmit>
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:365]: Here's how I think the `handleSubmit` logic should work.

I also flipped the ternary operators to avoid using a negative evaluation (`!== null`), which makes the code less brittle.

```suggestion
							<ButtonSubmit
								handleSubmit={messageIndexToEdit === null
									? handleSubmitNewMessage
									: handleSubmitEditMessage}
								hasMetaKey={isPromptFullscreen}
								disabled={!prompt ||
									!$settingsStore?.ollamaModels.length ||
									!$settingsStore?.ollamaModel}
							>
								{$i18n.t(
									messageIndexToEdit === null ? 'sessionsPage.run': 'saveAndRun'
								)}
							</ButtonSubmit>
```

fmaclen (changes requested):
The feature works as expected, aside from the **Cancel** button which shouldn't appear after re-submitting the message:

image

The rest of the comments are mostly minor code cleanups.
Once you are done with those we would need to add a new test for this behavior.

src/routes/sessions/[id]/+page.svelte
Comment (GregoMac1) [reply src/routes/sessions/[id]/+page.svelte:365]: I understand, but I added this reactive function because the old `handleSubmit` was not only used at the `ButtonSubmit`, but also at `handleKeyDown` and `FieldTextEditor`, so we would have that logic triplicated.

Comment (fmaclen) [reply src/routes/sessions/[id]/+page.svelte:352]: Oh, I see.. In that case can we keep `handleSubmit()` as it was and do the check for _Editing_ or _New_ inside of it once it gets called?

I'd like to avoid having a reactive function if possible.

Comment (GregoMac1) [reply src/routes/sessions/[id]/+page.svelte:352]: Sure! Is [this](https://github.com/fmaclen/hollama/pull/170/commits/e78d5477d157adaf10ee423b33f44aaf11addc84) what you mean?

src/routes/sessions/[id]/+page.svelte
Comment (GregoMac1) [reply src/routes/sessions/[id]/+page.svelte]: Note: Now that I've removed the `stream: true`, one of the tests is failing. I don't exactly see why, because when I run `npm run test` on my computer it works fine.

src/routes/sessions/[id]/+page.svelte
Comment (fmaclen) [reply src/routes/sessions/[id]/+page.svelte:352]: Yup, in general I try to avoid the reactive syntax unless we absolutely need to.

Also, not sure if you heard about the concept of ["Runes"](https://svelte-5-preview.vercel.app/docs/runes) coming up in Svelte 5... but I think that's going to require refactoring most (if not all) of our reactive declarations, I expect  that it's not going to be fun migrating from 4 to 5 😬

src/routes/sessions/[id]/+page.svelte
Comment (fmaclen) [reply src/routes/sessions/[id]/+page.svelte]: It's weird that it's working in your computer, but the test is failing because we it's expecting `stream: true` to be there but we just removed it, so it's good thing that the test caught it 👍 

image

In this case I think it's safe to remove it because Ollama's `stream` parameter defaults to `true`.
But just to double check make do some QA to make sure Ollama is indeed streaming a response back.

Comment (GregoMac1) [reply src/routes/sessions/[id]/+page.svelte]: Thank you! I've removed it from the test and it's working now.

tests/session.test.ts
```diff
@@ -482,6 +482,137 @@ test.describe('Session', () => {
...
+		);
+		await saveButton.click();
+		await expect(userMessage).toBeVisible();
+		await expect(userMessage).toContainText(
+			'Who would win in a fight between Scarlett Johansson and Jessica Alba?'
+		);
+		await expect(textEditorLocator(page, 'Prompt')).not.toBeVisible();
+		await expect(saveButton).not.toBeVisible();
+		await expect(cancelButton).not.toBeVisible();
+	});
```
fmaclen, line 522: Do we need this test? I think it has the same coverage as `can edit the first message and the next get removed` or am I missing something?

fmaclen (approved):
I think it's good to go 😎 👍
Left you a comment about a test that we can probably remove, let me know if I missed anything.

Also fixed a few small issues caused by my recent merge.

tests/session.test.ts
GregoMac1: I think you're right 😅. I'll remove it and merge this PR.

PR #174, opened 2024-09-05
chore: replace i18next with typesafe-i18n

fmaclen:
- Removes `i18next`
- Implements `typesafe-i18n`


.github/workflows/test.yml
```diff
@@ -2,7 +2,6 @@ name: Test
 
 on:
   pull_request:
-    branches: main
```
fmaclen, line 5: We want tests and linter to run on _any_ PR's, not just the ones based off of `main`.

package.json
```diff
@@ -5,7 +5,7 @@
 	"homepage": "https://hollama.fernando.is",
 	"description": "A minimal UI for talking to Ollama servers",
 	"scripts": {
-		"dev": "vite dev --host",
+		"dev": "concurrently \"typesafe-i18n\" \"vite dev --host\"",
```
fmaclen, line 8: `typesafe-i18n` has a "watcher" process that listens to changes in the locales and re-generates the type definitions, so I added the `concurrently` library to run both processes at the same time when running `npm run dev`.

src/lib/components/ButtonCopy.svelte
```diff
@@ -11,7 +11,7 @@
 </script>
 
 <div class="copy-button">
-	<Button title={$i18n.t('copy')} variant="icon" on:click={copyContent}>
+	<Button title={$LL.copy()} variant="icon" on:click={copyContent}>
```
fmaclen, line 14: With this library every key is a function you can call.

src/lib/components/SectionListItem.svelte
```diff
@@ -25,9 +25,7 @@
 		<a
 			class="section-list-item__a"
 			data-testid={isSession ? 'session-item' : 'knowledge-item'}
-			aria-label={(isSession
-				? $i18n.t('sessions', { count: 1 })
-				: $i18n.t('knowledge', { count: 1 })) + `: ${id}`}
+			aria-label={(isSession ? $LL.session() : $LL.knowledge()) + `: ${id}`}
```
fmaclen, line 28: We only really had one pluralized string: `Session/s`, while the library does [support pluralization](https://github.com/ivanhofer/typesafe-i18n/tree/main/packages/runtime#plural), I ended it up just adding dedicated keys for both versions.

src/routes/+layout.svelte
```diff
@@ -28,6 +24,11 @@
 		if (!($settingsStore.autoCheckForUpdates === false)) await checkForUpdates();
 	});
 
+	beforeUpdate(() => {
+		loadLocale('en');
+		setLocale('en');
+	});
```
fmaclen, line 30: You'll need to re-factor this for #151.
You can use my older `julia-sanfrancisco` repo [for](https://github.com/fmaclen/julia-sanfrancisco/blob/main/src/routes/%2Bpage.svelte#L41-L50) [reference](https://github.com/fmaclen/julia-sanfrancisco/blob/main/src/lib/player.ts#L62-L74).

src/routes/+layout.svelte
```diff
@@ -61,34 +62,37 @@
...
+					{$LL.knowledge()}
 				{:else if href === '/sessions'}
 					<MessageSquareText class="h-4 w-4" />
+					{$LL.sessions()}
 				{:else if href === '/settings'}
 					<Settings2 class="h-4 w-4" />
+					{$LL.settings()}
 				{:else if href === '/motd'}
 					<NotebookText class="h-4 w-4" />
+					{$LL.motd()}
```
fmaclen, line 84: Technically there is a way to call the locales dynamically but the docs weren't Svelte specific and since we are already doing a bunch of `if` statements for the icons, might as well use that to call the individual locales.

src/lib/components/FieldInput.svelte
```diff
@@ -1,9 +1,10 @@
 <script lang="ts">
 	import Field from './Field.svelte';
+	import type { LocalizedString } from 'typesafe-i18n';
 
 	export let name: string;
-	export let label: string;
-	export let placeholder: string | undefined = undefined;
+	export let label: LocalizedString;
+	export let placeholder: LocalizedString | string | undefined = undefined;
```
fmaclen, line 7: We now have a dedicated `LocalizedString` type, so unless we are expecting a random `string` we should use that type by default. The good thing is that the linter will yell at you if you do it wrong 😅

src/i18n/en/index.ts
```diff
@@ -0,0 +1,89 @@
+import type { BaseTranslation } from '../i18n-types';
+
+const en = {
+	allowConnections: 'Change your server settings to allow connections from',
+	allowMixedContent: 'Allow mixed content',
+	areYouSureYouWantToDeleteAll: 'Are you sure you want to delete all {type:string}?',
```
fmaclen, line 6: You can add type definitions to the interpolated variables, which is pretty nifty!

src/routes/settings/Ollama.svelte
```diff
@@ -125,42 +126,49 @@
...
+						{$LL.checkBrowserExtensions()}
 					</P>
 					{#if ollamaURL.protocol === 'https:'}
 						<P>
 							<!-- eslint-disable-next-line svelte/no-at-html-tags -->
-							{@html $i18n.t('tryingToConnectNotLocalhost')}
+							{@html $LL.tryingToConnectNotLocalhost({
+								hostname: '<code class="badge">localhost</code>',
+								ip: '<code class="badge">127.0.0.1</code>'
+							})}
```
fmaclen, line 160: Apparently there is a way to render components in the middle of a localized string but [we have to implement it ourselves](https://github.com/ivanhofer/typesafe-i18n/tree/main/packages/adapter-svelte#how-do-i-render-a-component-inside-a-translation). I don't think we should bother with that until we have many such cases.

In the meantime I figured it's best keeping all of our markup in the Svelte component and simply interpolate the HTML as a string. I don't think this approach will work in a lot of cases, but we'll cross that bridge when we have to.

.prettierignore
```diff
@@ -11,3 +11,10 @@ node_modules
 pnpm-lock.yaml
 package-lock.json
 yarn.lock
+
+# Ignore files for typesafe-i18n auto-generated files
+src/i18n/i18n-svelte.ts
+src/i18n/i18n-types.ts
+src/i18n/i18n-util.async.ts
+src/i18n/i18n-util.sync.ts
+src/i18n/i18n-util.ts
```
fmaclen, line 20: Prettier was auto-formatting these, but they will be re-generated every time we add/remove keys so we need to ignore them.

tests/settings.test.ts
```diff
@@ -36,9 +36,9 @@ test('handles server status updates correctly', async ({ page }) => {
 	await expect(page.getByLabel('Server')).toHaveValue('http://localhost:11434');
 
 	// The starting status is "connected"
-	await expect(page.getByText('disconnected')).not.toBeVisible();
-	await expect(page.getByText('connected', { exact: true })).toBeVisible();
-	await expect(page.getByText('connected', { exact: true })).toHaveClass(/badge--positive/);
+	await expect(page.getByText('Disconnected')).not.toBeVisible();
+	await expect(page.getByText('Connected')).toBeVisible();
+	await expect(page.getByText('Connected')).toHaveClass(/badge--positive/);
```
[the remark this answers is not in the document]
fmaclen: Fixed it!

`Connected`/`Disconnected` were not capitalized in the i18n locales, but they showed as capitalized in the UI because they were transformed by CSS. Got rid of all of that and did other cleanup work in that component while I was there.

playwright.config.ts
```diff
@@ -18,7 +18,7 @@ const config: PlaywrightTestConfig = {
 	snapshotPathTemplate: '{testDir}/{testFileDir}/{testFileName}-snapshots/{arg}{ext}',
 	expect: {
 		toMatchSnapshot: {
-			maxDiffPixelRatio: 0.01
+			maxDiffPixels: 500
```
fmaclen, line 21: I lowered the sensitivity of the snapshot tests because a lot of subtle UI changes were flying under the radar.

.prettierrc
```diff
@@ -20,6 +20,7 @@
 		"<BUILTIN_MODULES>",
 		"<THIRD_PARTY_MODULES>",
 		"",
+		"^([$]i18n)([/].*)$",
 		"^([$]env)([/].*)$",
 		"^([$]app)([/].*)$",
 		"^([$]lib)([/].*)$",
```
fmaclen, line 26: Added an import sorter: https://github.com/fmaclen/hollama/pull/176

PR #179, opened 2024-09-06
feat: translate UI to Spanish

GregoMac1:


GregoMac1:
@fmaclen for now I've added the translations to spanish. I still have to implement the switch between languages.

I've used Claude to give me advice on some of them. I still have doubts on a few ones, for example 'Prompt'. What do you think?

src/i18n/es/index.ts
```diff
@@ -0,0 +1,91 @@
+import type { BaseTranslation } from '../i18n-types';
+
+const es = {
+	allowConnections: 'Cambia la configuración de tu servidor para permitir conexiones desde',
+	allowMixedContent: 'Permitir contenido mixto',
+	areYouSureYouWantToDeleteAll: '¿Estás seguro de que quieres eliminar todos los {type:string}?',
+	areYouSureYouWantToLeave: '¿Estás seguro de que quieres salir?\nLa tarea en progreso se detendrá',
```
fmaclen, line 7: Me suena raro "de que quieres", no?

```suggestion
	areYouSureYouWantToDeleteAll: '¿Estás seguro que quieres eliminar todos los {type:string}?',
	areYouSureYouWantToLeave: '¿Estás seguro que quieres salir?\nLa tarea en progreso se detendrá',
```

fmaclen (commented):
@GregoMac1 this translations look good, the one for _prompt_ is kinda weird but it's fine for now.

I edited the original issue and added a short list of tasks we should include in this PR:
https://github.com/fmaclen/hollama/issues/151

src/i18n/es/index.ts
fmaclen: No se por que me suena raro cuando lo leo la primera vez, pero me suena bien cuando lo leo una segunda vez 🤷 
Igual ok, que quede como esta 👍

GregoMac1:
@fmaclen as the settings page is divided into categories, I added a 'Preferences' category to contain the language switch input, but feel free to change/remove it.

src/routes/+layout.svelte
```diff
@@ -25,12 +27,13 @@
...
 	});
 
-	beforeUpdate(() => {
-		loadLocale('en');
-		setLocale('en');
-	});
-
 	onMount(() => {
+		if (!$settingsStore.userLanguage)
+			$settingsStore.userLanguage = detectLocale('en', ['en', 'es']) as Locales;
```
GregoMac1, line 32: `detectLocale` accepts a fallback language as first parameter, and a list of available locales as second parameter. [Ref](https://github.com/ivanhofer/typesafe-i18n/tree/main/packages/detectors#detectors)

fmaclen (changes requested):
This is pretty good so far 👍 
Left you a couple of minor changes.

Also pushed a few styling fixes because in Spanish strings are often much longer and the UI was wrapping weirdly.

Last but not least, we needs some tests:
- User can set the language preference to `es`, and back to `en`.
- If the user's browser is in `es` make sure that's loaded by default, I think you can achieve this by [emulating the test options](https://playwright.dev/docs/test-use-options#emulation-options), thought instead of defining a config we should [`test.use()`](https://playwright.dev/docs/api/class-test#test-use)
- Update the snapshots `npx playwright test --update-snapshots`

GregoMac1:
@fmaclen I've added the tests you mentioned.

Could you please try updating the snapshots? I've already done so but the test is still failing. Thanks!

fmaclen:
> Could you please try updating the snapshots?

Done!

src/routes/settings/Language.svelte
```diff
@@ -0,0 +1,33 @@
...
+	<FieldSelect
+		name="language"
+		label={$LL.language()}
+		bind:value
+		options={[
+			{ value: 'en', option: 'English' },
+			{ value: 'es', option: 'Español' }
+		]}
+	/>
+</Fieldset>
```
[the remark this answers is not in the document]
fmaclen: @GregoMac1 we still need to do the first part of this comment which is setting the default `value`.

And it'd probably be good to add assertions for the default values in both of the tests.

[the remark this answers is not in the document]
fmaclen: I now realize that my comment was a bit ambiguous, I meant not to worry about fixing the hardcoded empty value, but we still need a default value 😅

GregoMac1:
@fmaclen I've updated the test, so I think this would be ready to be merged.

However, the snapshots test is still failing. I updated the snapshots and the test passes on my computer, but the GH Action fails. I don't really know how to debug this further.

fmaclen (approved):
LGTM 🚀

fmaclen:
> the snapshots test is still failing

@GregoMac1 I think the issue with the snapshots is still indeed the antialias and how sensitive the config is:

```
docs.test.ts:15:1 › seed data and take screenshots for README.md 
Error: Screenshot comparison failed:
863 pixels (ratio 0.01 of all image pixels) are different.
```

I increased it to 900 pixels, hopefully this is strikes a good balance for catching actual UI changes and not false positive.

GregoMac1:
Awesome! Thanks!

Issue #94, opened 2024-07-18
Support OpenAI API format


anagri:
Support OpenAI API format by giving option to switch between Ollama proprietary API format and OpenAI API format.

To fetch list of models - https://platform.openai.com/docs/api-reference/models/list?lang=node.js

To generate chat - https://platform.openai.com/docs/api-reference/chat?lang=node.js

Will help servers following OpenAI API format be used with Hollama. E.g. Bodhi App - https://github.com/BodhiSearch/BodhiApp

fmaclen:
I'm not sure entirely sure I understand this request, or what the use-case would be.

It sounds like you want Hollama to expose an API using the OpenAI schema so other services can interact with Ollama using Hollama as a proxy of sorts? is that correct?

binarynoise:
I understand that he rather wants to connect to OpenAI-API-compatible servers that do not use the ollama-API.

fmaclen:
@binarynoise What are "OpenAI-API-compatible servers"?

Is the suggested feature for adding a way to set your own OpenAI API key and be able to choose to get a completion from Ollama or OpenAI's API?

binarynoise:
> What are "OpenAI-API-compatible servers"?

Bodhi App seems to be one:

> It also exposes these LLM inference capabilities as OpenAI API compatible REST APIs.

@anagri would have to clarify the rest.

anagri:
tx @binarynoise 

@fmaclen - many of the apps that run LLMs locally prefer exposing APIs using OpenAI API format. These includes jan.ai, lm-studio, bodhi app etc.

so if hollama can start supporting OpenAI APIs, we can use hollama UI to connect to the APIs exposed by these apps. and if someone wants to use their OpenAI API key, they can use that as well.

tldr; - OpenAI APIs are more of industry standard than proprietary Ollama APIs. Hollama get can a wider audience and adoptability if they start supporting OpenAI APIs.

fmaclen:
@binarynoise @anagri thanks for the clarification, I understand what you mean now.

fmaclen:
To expand a bit, I'm not opposed to the idea of allowing multiple completion servers, this is however not a trivial implementation and we have other priorities at the moment.

If someone wants to give this a shot I'd start by using [`ollama.ts`](https://github.com/fmaclen/hollama/blob/main/src/lib/ollama.ts) as the base template to handle the logic of the other servers.

There are also a few factors to consider for this implementation:
[...]

PR #186, opened 2024-09-16
feat: sort models list by "the last 5 used", then alphabetically

### LINKED ISSUE #143: Refactor `<select>` menus so we can have grouped options and fuzzy search
REF #124

GregoMac1:
Also closes #143 

image


GregoMac1:
@fmaclen I forgot adding the `<Badge />` next to each model, you don't have to review this yet 😅

src/lib/components/FieldSelect.svelte
```diff
@@ -97,15 +120,37 @@
...
+							<Check class="base-icon" />
+						</Combobox.ItemIndicator>
+						<div class="field-combobox-item-label">
+							{group.label}
+							{#if group.extra}
+								<Badge>{group.extra}</Badge>
+							{/if}
+						</div>
+					</Combobox.Item>
+				{/if}
```
GregoMac1, line 153: @fmaclen maybe we could extract `<Combobox.Item>` to a separate component so we remove these duplicated lines, but I don't know if it's worth it.

src/lib/sessions.ts
```diff
@@ -76,3 +76,15 @@ export function getSessionTitle(session: Session) {
...
+	const currentSessions = get(sessionsStore) || [];
+	const lastUsedModels: string[] = [];
+
+	for (const session of currentSessions) {
+		if (!lastUsedModels.includes(session.model)) lastUsedModels.push(session.model);
+		if (lastUsedModels.length >= 5) break;
+	}
+
+	return lastUsedModels;
+}
```
fmaclen, line 90: We are only using this function in `FieldSelectModel` so let's keep this logic there, maybe rename it to `handleRecentlyUsedModels` so it's consistent with `handleChange()` and the label on the screen.

If we find we need to use this in other places later we might end up moving it back to `sessions.ts`.

src/lib/components/FieldSelectModel.svelte
```diff
@@ -3,13 +3,23 @@
...
+	$: models = $settingsStore.ollamaModels.map((m) => ({
+		value: m.name,
+		label: m.name,
+		extra: m.details.parameter_size
+	}));
+	$: lastUsedModels = getLastUsedModels().map((m) => ({
+		value: m,
+		label: m,
+		extra: models.find((model) => model.value === m)?.extra
+	}));
```
fmaclen, line 22: - [ ] If there are no recent models it renders an empty list. If a group doesn't have any results we should include an empty message: `No recently used models`. I thought of suggesting hiding the group entirely but I feel like it's better UX if the user knows the feature even exists before they use a model for the first time.
  image

- [ ] The models in the recent list are duplicated in the other list, so when you type a model name you see the result twice. We probably shouldn't include the model in the 2nd group.
  image

- [ ] Alphabetical sorting got lost in the first group:
  image

src/lib/components/FieldSelect.svelte
fmaclen: Yeah, we could DRY the `<Combobox.Item>` but I think your current version looks more readable. Let's keep it 👍

src/i18n/en/index.ts
```diff
@@ -1,6 +1,7 @@
 import type { BaseTranslation } from '../i18n-types';
 
 const en = {
+	allModels: 'All models',
```
fmaclen, line 4: Let's change this to `All other models` so the description matches the new behavior.

src/lib/components/FieldSelect.svelte
```diff
@@ -151,4 +196,12 @@
 	:global(.field-combobox-item-label) {
 		@apply w-full;
 	}
```
fmaclen, line 198: Here are some styles for ya:

```suggestion
	:global(.field-combobox-item-label) {
		@apply w-full grid grid-cols-[auto,max-content];
	}
```

image

src/lib/components/FieldSelect.svelte
```diff
@@ -4,34 +4,55 @@
...
 	export let disabled: boolean | undefined = false;
-	export let options: Selected<string>[] = [];
+	export let options: OptionOrGroup[] = [];
 	export let value: string | undefined = undefined;
 	export let placeholder: string = '';
-	export let onChange: (value: Selected<string>) => void = () => {};
+	export let onChange: (value: Option) => void = () => {};
 
-	const noSelection = { value: '', label: '' };
+	type Option = Selected<string> & { extra?: string };
```
fmaclen, line 19: Let's rename `extra` > `badge` so it's easier to _guess_ what component the value will be rendered in.

src/lib/components/FieldSelect.svelte
```diff
@@ -51,8 +72,10 @@
 
 	function handleClear() {
 		value = undefined;
-		selected = noSelection;
-		onChange(noSelection);
+		requestAnimationFrame(() => {
+			selected = noSelection;
+			onChange(noSelection);
+		});
```
fmaclen, line 78: Why was `requestAnimationFrame` needed?

fmaclen (changes requested):
Man, I've been looking forward to this feature!

PR looks good so far. Left you some minor issues and a CSS suggestion to fix the styling a bit. Needs some tests 🧪

src/i18n/es/index.ts
```diff
@@ -78,6 +80,8 @@ const es = {
 	run: 'Ejecutar',
 	save: 'Guardar',
 	saveAndRun: 'Guardar y ejecutar',
+	search: 'Buscar',
+	searchEmpty: 'No se encontraron resultados',
```
fmaclen, line 84: Were these missing from my PR?

I think we need to add a CI check for this i18n stuff, it's easy to forget to translate stuff and/or re-generate the types.

src/lib/components/FieldSelect.svelte
GregoMac1: I made it look like this at the beginning because it looks good on the New session screen, but in the Settings screen I'm not sure if it looks nice because the input is wider. What do you think?

image

src/lib/components/FieldSelectModel.svelte
GregoMac1: > Alphabetical sorting got lost in the first group

Actually, I did this on purpose. I thought it'd be better to have the recently used models sorted by last used. Then, in the second group, models are sorted alphabetically. Should both groups be sorted alphabetically?

fmaclen: Ah, interesting. I'm not sure what the best sorting way should be, I'll have to to play with it.
Let's keep your current sorting by frequency, sounds like a good bet.

Make sure we add a couple of assertions for that sorting logic.

src/i18n/en/index.ts
GregoMac1: Okay, so the recently used models shouldn't be included in the second group of models?

fmaclen: Correct

GregoMac1:
All done! I still have to add some tests.

src/lib/components/FieldSelectModel.svelte
fmaclen: Maybe we should reverse the order, in this screenshot you can see 2 of the 3 sessions with `codestral:latest` but `phi:latest` appears first:

image

fmaclen: I see, for some reason I always pictured this feature with the most popular at the top.
But that's okay, let's leave it with the current order.

GregoMac1: We could also sort the models by _most used_, but there'd be some details that we should consider:
- we should iterate over the sessions and count how many times was each model used. Then, we could sort the 5 most used models. However, if the user had a large number of sessions, we should see if this still performs well.
- let's suppose that my 5 most used models have ~100 sessions each. If I pulled some new models and preferred to use them from now on, they wouldn't appear in the first group of models until I have more sessions that use those newly pulled models.

fmaclen: You are right, there's quite a bit of complexity if we go the _most used_ route.
What we have now is already a huge UX improvement, we can punt these ideas for later.

src/lib/components/FieldSelect.svelte
```diff
@@ -149,6 +204,18 @@
 	}
 
 	:global(.field-combobox-item-label) {
-		@apply w-full;
+		@apply grid w-full cursor-pointer grid-cols-[auto,max-content] gap-x-1;
+	}
+
+	:global(.field-combobox-item-label-option) {
+		@apply overflow-hidden text-ellipsis text-nowrap;
+	}
```
fmaclen, line 212: Fixed the styling for mobile:

image

tests/settings.test.ts
```diff
@@ -203,3 +209,95 @@ test.describe('locales', () => {
...
+
+	await modelComboBox.fill('nonexistentmodel');
+	await expect(page.getByText('No results')).toBeVisible();
+});
+
+test('models are correctly grouped including recently used models', async ({ page }) => {
+	await page.goto('/settings');
+	await mockTagsResponse(page);
+	await chooseFromCombobox(page, 'Available models', MOCK_API_TAGS_RESPONSE.models[1].name);
+
```
fmaclen, line 239: There is some duplication of tests with [`ui.test.ts`](https://github.com/fmaclen/hollama/blob/main/tests/ui.test.ts)
I think it makes more sense to move the non-duplicated tests to that file.

You can add a new `test.describe('FieldSelectModel', () => { ... });` section for the new tests.

Also, let's test these features in a the **New session** section, for example: `await page.goto('/sessions/qbhc0q');`, because we are likely to remove the models menu from the **Settings** page after #94

fmaclen (changes requested):
Almost there, only minor code organization changes.

tests/ui.test.ts
```diff
@@ -66,4 +79,51 @@ test.describe('FieldSelect', () => {
 		await expect(page.getByRole('option')).toHaveCount(0);
 		await expect(page.getByTitle('Clear')).not.toBeVisible();
 	});
+
+	test('models are correctly grouped and sorted', async ({ page }) => {
+		const newSessionButton = page.getByTestId('new-session');
+		await page.goto('/settings');
+		await mockTagsResponse(page);
```
GregoMac1, line 86: I had to do this because If I did `await page.goto('/sessions/qbhc0q');` instead of going to `/settings` first, the `Run` button would never get enabled.

tests/ui.test.ts
```diff
@@ -66,4 +79,51 @@ test.describe('FieldSelect', () => {
...
+		const recentlyUsedModels = recentlyUsedModelsGroup.locator('.field-combobox-item-label-option');
+		await expect(recentlyUsedModels).toHaveCount(2);
+		await expect(recentlyUsedModels.nth(0)).toHaveText(MOCK_API_TAGS_RESPONSE.models[0].name);
+		await expect(recentlyUsedModels.nth(1)).toHaveText(MOCK_API_TAGS_RESPONSE.models[1].name);
+
+		// Check for "Other models" group
+		const otherModelsGroup = page.locator('.field-combobox-group', {
+			hasText: 'Other models'
+		});
+		await expect(otherModelsGroup).toContainText(MOCK_API_TAGS_RESPONSE.models[2].name);
```
fmaclen, line 127: Let's add one minor assertion here to make sure the models are not duplicated across both groups.

```suggestion
		await expect(otherModelsGroup).toHaveCount(1);
		await expect(otherModelsGroup).toContainText(MOCK_API_TAGS_RESPONSE.models[2].name);
```

fmaclen (approved):
This looks great! 😎 
I left you a minor assertion suggestion but feel free to merge this after adding it.

tests/ui.test.ts
GregoMac1: For some reason the `Run` button is still disabled when the tests are run in the GH Action. Debugging...

fmaclen: Gotcha, yeah, let's get that sorted out before merging.

GregoMac1: All tests pass if I set `retries: 2` in `playwright.config.ts` 🤔

PR #188, opened 2024-09-17
chore: update label for submit prompt button

fmaclen:


src/routes/sessions/[id]/+page.svelte
```diff
@@ -339,11 +339,7 @@
 									$settingsStore.ollamaModels.length === 0 ||
 									!$settingsStore.ollamaModel}
 							>
-								{#if messageIndexToEdit}
-									{$LL.run()}
-								{:else}
-									{$LL.saveAndRun()}
-								{/if}
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:346]: This logic was backwards, but also when `messageIndexToEdit === 0` it came up as _falsy_ so it never worked.
That being said, I don't think there is real benefit in having 2 distinct labels so I'm removing it.

PR #190, opened 2024-09-19
chore: generate snapshots in CI

### LINKED ISSUE #189: Tests: switch runner to `ubuntu-latest` and generate snapshots in CI

fmaclen:
Closes #189

- Integration tests now run on `ubuntu-latest`.
- We run Playwright with the `--update-snapshots` flag.
- If the UI is updated it will automatically generate the new snapshot and if the tests passed a Github bot will commit those new snapshots to the PR, that way we can review UI changes together with the code.
- Also added a new [`--reporter=list`](https://playwright.dev/docs/test-reporters#list-reporter) flag so we can see the status of each test as they run in the CI logs.
- Updated `package.json` so when we run `npm test` locally we use the `--ignore-snapshots` flag, otherwise our local tests will always fail due to the differences between macOS and Linux renderers.

This workflow may have have some unexpected behaviors but I'm willing to give it a shot and see how it plays out. One, potentially annoying, thing is that after you commit your changes to the branch, you will then have to pull the new snapshots once the tests are finished.


tests/docs.test.ts-snapshots/session.png
fmaclen, line 1: Sadly the font rendering in Linux didn't come out as nice as macOS, it appears that the `letter-spacing` is off by a few pixels. I think that's okay for now, we can investigate ways of improving that later.

GregoMac1: I agree. When you compare the snapshots side to side, they look better on MacOS, but I think they still look good on ubuntu.

GregoMac1 (approved):
This looks great! I think it'll be a great testing experience improvement 🚀

PR #194, opened 2024-09-22
feat(i18n): Add Turkish translation

myteril:
This pull request adds Turkish language support to the app. 

I have implemented the following changes.
- Translated all strings into Turkish.
- Created new translation file `src/i18n/tr/index.ts`.
- Ran the command `typesafe-i18n` to ensure the new file is integrated with the app.
- Added Turkish to the language list in the settings.
- Tested the app and confirmed functionality (see the screenshot below).

I hope it helps make the app more accessible to Turkish speakers.

PR Screenshot


fmaclen (changes requested):
Thanks for submitting this PR 👍 
I think this looks good, couple of minor changes:

1- We need a simple test that checks the language switches to Turkish, I think the easiest way to do it is to essentially duplicate [this test for Spanish](https://github.com/fmaclen/hollama/blob/main/tests/settings.test.ts#L188-L204), just update the browser locale accordingly.
2- Looks like there are some [linting errors](https://github.com/fmaclen/hollama/actions/runs/10982606989/job/30587311793?pr=194), you can run `npm run format && npm run lint` on your machine to fix them automatically or get more details.

myteril:
In the last commit, I added the test case and fixed linting. Could you please review it again?

fmaclen (approved):
@myteril thanks for adding the test. I had to make some changes to it because our existing test setup was only meant for Spanish.

I also spotted some issues with some recent changes we made to our test runner but we'll get that fixed later.

Anyways, this is good to go 👍

PR #196, opened 2024-09-23
fix: disallow clearing the language input

GregoMac1:
Added an `allowClear` prop to `FieldSelect`, which defaults to true but is set to false in this case.


fmaclen (approved):
LGMT 😎

PR #202, opened 2024-09-30
feat: add support for OpenAI

GregoMac1:


src/lib/localStorage.ts
```diff
@@ -51,10 +51,12 @@ export enum StorageKey {
 }
 
 export interface Settings {
+	apiType: 'ollama' | 'openai';
 	ollamaServer: string | null;
 	ollamaModel: string | null;
 	ollamaModels: ModelResponse[];
 	ollamaServerStatus: 'connected' | 'disconnected';
```
fmaclen, line 58: I don't think we want to store an `apiType` in settings, in fact, I think we should get rid of `ollamaModel` as well because after #193 you can switch models at any time.

The chosen model is what's going to determine if we hit the Ollama or OpenAI APIs.

By default the chosen model would be whichever is the last one you used for the **current session**.

Then, when you start a **new session** the default model should be the one at the top of the list in `Recently used`:
image

Or no model selected if there are no sessions.

fmaclen (changes requested):
I haven't had time to do a lot of deep analysis but directionally I think it's a great start 👍 

Left you some comments related to how I think we should choose one strategy over the other.

Another minor thing is that we can probably drop the `strategy` naming convention, at least on the file names.

GregoMac1:
Thanks for the feedback! I'm working on it 🫡

src/lib/chat/index.ts
```diff
@@ -0,0 +1,34 @@
...
+
+function getChatStrategy(): ChatStrategy {
+	const settings = get(settingsStore);
+	if (!settings) throw new Error('No settings specified');
+
+	if (settings.apiType === 'ollama') {
+		return new OllamaStrategy();
+	} else if (settings.apiType === 'openai') {
+		return new OpenAIStrategy();
+	}
```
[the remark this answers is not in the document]
GregoMac1: Nice, I agree with this.
Now my question is how to handle the list of available models. If we want to allow the user to switch from an ollama model to a openai model during a session, then 'available models' should be the combination of the models of each strategy. Am I getting it right?

GregoMac1:
@fmaclen Just to be sure: should we get rid of all the 'connected' or 'disconnected' logic? For example, in the `settings` page, should the status badge be removed?

fmaclen:
> should we get rid of all the 'connected' or 'disconnected' logic? 

@GregoMac1 let's keep that functionality only in the Settings view, it's still useful to know if we are correctly connected to an Ollama server.

GregoMac1:
Remaining tasks:
- [ ] Match types between both APIs.
- [ ] Check if the OpenAI strategy works correctly with a real API key.
- [ ] Add missing tanslations for `es`, `ja`, and `tr`.

src/routes/settings/OpenAI.svelte
```diff
@@ -0,0 +1,87 @@
...
+
+<Fieldset>
+	<P><strong>OpenAI</strong></P>
+	<FieldInput
+		name="baseUrl"
+		label={$LL.baseUrl()}
+		placeholder={DETAULT_OPENAI_SERVER}
+		bind:value={openaiServer}
+		on:keyup={getModelsList}
+	>
```
fmaclen, line 55: First, in `onMount`:
- let's remove the `getModelsList` call when the component mounts.
- `openAiURL` should be `https://api.openai.com/v1`, not `localhost`.

Then, one thing I noticed is that after successfully connecting to OpenAI, if I leave away from **Settings** and then come back, it's always showing as `disconnected` first.

I understand that this is how the Ollama implementation was working but now that the connection to OpenAI is much slower it's very noticeable and feels like mistake.

I'm wondering if we should even remove the `onKeyUp` events and add a button to intentionally "Connect" to the server `onClick`, similar to the one for pulling Ollama models.

If we add a button then we can probably remove the `connected/disconnected` badges and simply show a `Connecting`, `Connected/Disconnected` toast alerts (or maybe even something more useful like status updates, for example: `Retrieving models`).

Anyways, let's not add a button yet, I'm just thinking out loud.

Let's fix all of the other stuff first and we'll come back to this later once we have a better feel for the rest of the UX.

fmaclen (changes requested):
Wasn't able to try the functionality, when visiting a Session I kept getting a `Can't connect to Ollama server` error, even though my Ollama server was running.

Left you several comments but primarily I think we should limit the amount of calls we do to the API.

src/routes/sessions/[id]/+page.svelte
```diff
@@ -73,9 +73,9 @@
 
 	async function handleSessionChange() {
 		try {
-			$settingsStore.ollamaModels = (await ollamaTags()).models;
+			$settingsStore.models = await listModels();
 		} catch {
-			$settingsStore.ollamaModels = [];
+			$settingsStore.models = [];
 			toast.warning($LL.cantConnectToOllamaServer());
 		}
```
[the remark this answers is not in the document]
Comment (GregoMac1) [reply src/routes/sessions/[id]/+page.svelte]: I've been thinking about this. Wouldn't it be better to request the models in the Sessions page? Because for example if I open Hollama and go to Sessions instead of going to Settings first, then the models wouldn't be available when I click 'New session'.

src/i18n/es/index.ts
```diff
@@ -57,16 +61,31 @@ const es = {
 	language: 'Idioma',
 	lastUsedModels: 'Modelos utilizados recientemente',
 	light: 'Claro',
+	lowVram: 'Bajo VRAM',
+	mainGpu: 'GPU principal',
```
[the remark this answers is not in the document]
GregoMac1: I agree. I've translated them because I wanted to keep the UI's language consistent, but it's true that these terms are usually shown in english no matter the language of the speaker.

src/routes/settings/OpenAI.svelte
GregoMac1: Thanks for the feedback! I've improved several things about the OpenAI config, please let me know what you think

package.json
```diff
@@ -67,6 +68,8 @@
 	},
 	"optionalDependencies": {
 		"@cloudflare/workerd-linux-64": "^1.20240909.0",
+		"@esbuild/darwin-x64": "^0.24.0",
+		"@esbuild/win32-x64": "^0.24.0",
```
fmaclen, line 72: Why were these needed?

src/lib/components/FieldInput.svelte
```diff
@@ -47,7 +47,7 @@
 			id={name}
 			{disabled}
 			{placeholder}
-			type="text"
+			{...{ type }}
```
fmaclen, line 50: Never seen this syntax, can you do `{type}` directly?

src/routes/sessions/[id]/+page.svelte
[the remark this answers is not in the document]
Comment (fmaclen) [reply src/routes/sessions/[id]/+page.svelte]: Yeah, it would be nicer to check for models when switching Sessions, but the reality is that models don't change that often and hitting the API any time `handleSessionChange` felt like we were sending way to many requests.

A more "proper solution" could be to check the models `handleSessionChange` with a debounce so we only check for models every 10 minutes or something like that. We have similar logic for [checking for version updates](https://github.com/fmaclen/hollama/blob/main/src/lib/updates.ts#L50-L54) once a week but I feel like that might add too much complexity to this PR, that's why I was suggesting keeping it only in Settings for now, and improve it later.

src/routes/settings/OpenAI.svelte
```diff
@@ -0,0 +1,86 @@
...
+		<svelte:fragment slot="nav">
+			<Button
+				aria-label="Connect"
+				class="h-full text-muted"
+				isLoading={openaiServerStatus === 'connecting'}
+				disabled={openaiServerStatus === 'connecting' || !openaiServer || !openaiApiKey}
+				on:click={updateOpenAIConfig}
+			>
+				<MonitorUp class="base-icon" />
+			</Button>
```
fmaclen, line 67: - Lets change the icon to `<RefreshCw />` so it's more aligned to the behavior, because we are basically "syncing" with OpenAI.
- We should use the secondary button style, otherwise it "stays orange" after a successful connection
- `aria-label` is missing i18n key

src/routes/settings/OpenAI.svelte
```diff
@@ -0,0 +1,86 @@
...
+	let openai = new OpenAIStrategy();
+	let openaiServer = $settingsStore.openaiServer || DEFAULT_OPENAI_SERVER;
+	let openaiApiKey = $settingsStore.openaiApiKey || '';
+	let openaiServerStatus: 'connected' | 'disconnected' | 'connecting' = 'disconnected';
+
+	$: settingsStore.update((settings) => ({ ...settings, openaiServer, openaiApiKey }));
+
+	async function getModelsList(): Promise<void> {
+		openaiServerStatus = 'connecting';
+		toast.warning($LL.connecting());
```
fmaclen, line 25: Let's use the "loading" toast instead of warning, [here's an example](https://github.com/fmaclen/hollama/blob/main/src/routes/settings/Ollama.svelte#L53).

src/lib/components/FieldInput.svelte
GregoMac1: I tried that first but I kept getting [this error](https://stackoverflow.com/questions/57392773/error-type-attribute-cannot-be-dynamic-if-input-uses-two-way-binding). I could only solve it by applying that syntax, which is described in [this reply](https://stackoverflow.com/questions/57392773/error-type-attribute-cannot-be-dynamic-if-input-uses-two-way-binding/75298645#75298645). I left the reference in the [commit](https://github.com/fmaclen/hollama/pull/202/commits/b332a85a0cac75ce638c9af003a775578f027297).

src/routes/sessions/[id]/+page.svelte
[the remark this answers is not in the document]
Comment (GregoMac1) [reply src/routes/sessions/[id]/+page.svelte]: I'm sorry, I didn't mean to retrieve the models when switching sessions.

As far as I understand, that would happen if we retrieve them in `src/routes/sessions/[id]/+page.svelte`. What I am saying is that we should do it in `src/routes/sessions/+page.svelte`. That way, we only retrieve the models when going to `/sessions`.

I think this is approach is better than retrieving the models in `/settings` because of the following case:
If I delete all my settings from localStorage, then `$settingsStore.models` is empty. If I enter Hollama and go to `/sessions` directly before going to `/settings`, then the models should be available so I can start a new session.

package.json
GregoMac1: I was having errors when running `npm run dev` on both of my computers (with macOS x64 and Windows 11 64). Now, in macOS, I removed the `@esbuild/darwin-x64` package and it is still working 🤔

src/lib/components/FieldInput.svelte
fmaclen: Understood, let's add that reference as a comment in the code. I would have never looked at the commit comment 😅

src/routes/sessions/[id]/+page.svelte
```diff
@@ -137,9 +129,14 @@
...
-				$editor.completion += chunk;
-				await scrollToBottom();
+			await chat({
+				model: $settingsStore.models.find((model) => model.name === $session.model) as Model,
+				payload: ollamaChatRequest,
+				abortSignal: $editor.abortController.signal,
+				onChunk: async (chunk) => {
+					$editor.completion += chunk;
+					await scrollToBottom();
+				}
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:139]: I came across some issues with `try/catch { }` by turning off the Ollama server or shutting off my WiFi to trigger a network error. Here's what happens when I turn off WiFi and attempt to submit the prompt to OpenAI:

Screenshot 2024-10-16 at 11 57 01 AM

That being said, since that's an existing bug I added a new ticket to do it separately.
https://github.com/fmaclen/hollama/issues/218

On a side note, are you generally doing QA on these kinds of edge cases?

For example, another thing that I haven had time to try but would be useful is to intentionally edit the API key in `localStorage` and try to submit a prompt with an _invalid key_. I expect that to more or less behave like the network error but we need to make to make sure it fails gracefully with our existing implementation.

src/routes/+layout.svelte
```diff
@@ -68,6 +68,7 @@
 		classes: {
 			toast:
 				'shadow-xl px-4 py-3 flex items-center gap-x-3 max-w-full w-full rounded mx-auto text-xs mx-0',
+			loading: 'text-shade-12 bg-shade-4',
```
fmaclen, line 71: This needs the same changes as PR #217

fmaclen (changes requested):
I think this is more or less there, just left you some minor comments below.
Let's get those done and get started with the tests.

src/routes/sessions/[id]/+page.svelte
Comment (GregoMac1) [reply src/routes/sessions/[id]/+page.svelte:139]: Nice, good catch!

I've just tried to edit the API key in the localStorage. The behavior is the following:
- If I edit the API key in an ongoing session, I can keep chatting because I think the API key is already stored at RAM memory.
- If I do the same and I refresh the page, then it shows the error that comes from OpenAI:
image

fmaclen (changes requested):
I think this looks good.
Let's add the tests next.

src/lib/components/FieldSelect.svelte
```diff
@@ -144,7 +144,17 @@
 												{option.label}
 											</span>
 											{#if option.badge}
-												<Badge>{option.badge}</Badge>
+												{#if Array.isArray(option.badge)}
+													<div class="field-select-badge">
+														{#each option.badge as badge}
+															<Badge variant={badge === 'OpenAI' ? 'openai' : badge === 'Ollama' ? 'ollama' : undefined}>
+																{badge}
+															</Badge>
```
fmaclen, line 152: Added icons for Ollama and OpenAI.

image

src/routes/sessions/[id]/+page.svelte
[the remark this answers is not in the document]
Comment (fmaclen) [reply src/routes/sessions/[id]/+page.svelte]: Sorry, I missed your question. I like your suggestion of doing the check in `src/routes/sessions/+page.svelte` 👍

src/lib/components/FieldInput.svelte
fmaclen: Don't forget to add this comment to the codebase

fmaclen:
> The only bug I found was that after choosing an OpenAI model (i.e. gpt-3.5-turbo) and submitting a prompt, I was expecting it to show up in the Recently used models list. It did show up later after I left the Sessions page and returned to it.

@GregoMac1 I'm still seeing this issue.

src/lib/chat/openai.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export class OpenAIStrategy implements ChatStrategy {
+	private openai: OpenAI;
+
+	constructor() {
+		const settings = get(settingsStore);
+
+		this.openai = new OpenAI({
+			baseURL: settings.openaiServer,
+			apiKey: settings.openaiApiKey || '',
```
fmaclen, line 16: There is an issue with this approach, we shouldn't instantiate a `new OpenAI()` object without an `apiKey`.

Looks like `|| ''` was added to satisfy the TS linter, but his code is currently causing `401` responses where there is no `apiKey` defined in Settings (which is the default behavior):

image

fmaclen: Likewise we also shouldn't instantiate `new OpenAI` without a `openaiServer`

src/routes/sessions/[id]/+page.svelte
```diff
@@ -47,7 +45,6 @@
 	let messagesWindow: HTMLDivElement;
 	let userScrolledUp = false;
 
-	$: $editor.isNewSession = !$session?.messages.length;
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:50]: Not entirely sure what the issue was but I moved this line to [`handleSessionChange()`](https://github.com/fmaclen/hollama/pull/202/files#diff-e13c06b2f8eb5ca2f92007ee68d5fecf31e07131e5f14dcfb9a58c00efc01fe3R74) and it fixed #213.

src/lib/chat/openai.ts
fmaclen: Maybe instead of reading `openAiApiKey` and `openAiServer` from the store, maybe they should be required arguments, for example: `new OpenAIStrategy(server: string, apiKey: string)`.

src/routes/sessions/[id]/+page.svelte
Comment (fmaclen) [reply src/routes/sessions/[id]/+page.svelte:140]: I moved all the errors in Sessions to be toasts in #219
Also, merged those changes into this branch and settled the conflicts.

Works for trying to hit OpenAI without Internet:

image

And with incorrect credentials:

Screenshot 2024-10-19 at 8 42 31 AM

It's a little ugly but at least is informative and you'll likely never see these errors.

fmaclen:
Here's a test of tests cases:
- Fetching data from OpenAI with a correct API Key
- Trying to fetch data with an incorrect API key
- Trying to fetch data with a network connection error
- Can't send any fetch requests without an `apiKey` or `baseUrl` set
- Check the sorting/order of the models list is correct
- Once an OpenAI model is used it gets added to the "recently used" list
- An OpenAI model can be used in a session and that the model is saved to the localStorage for that specific session
- In `FieldModelSelect`:
  - Check only the `gpt` filtered models are available 
  - Check that Ollama models have an `ollama` badge
  - Check that OpenAI models have an `openai` badge

GregoMac1:
> > The only bug I found was that after choosing an OpenAI model (i.e. gpt-3.5-turbo) and submitting a prompt, I was expecting it to show up in the Recently used models list. It did show up later after I left the Sessions page and returned to it.
> 
> @GregoMac1 I'm still seeing this issue.

@fmaclen This should be fixed now!

fmaclen (changes requested):
I pushed some changes for updating the `MOTD` and `README`.
I also merged the latest set of changes from `main` and now there are 3 test failures.

Everything else looks good AFAIK, let's get those test fixed and we'll get this out 🚀

fmaclen:
@GregoMac1 I get these 2 tests failing:

```
openai.test.ts:56:2 › OpenAI Integration › models list is sorted correctly
session.test.ts:720:2 › Session › handles errors when fetching models
```
---

This one I'm not entirely sure why it's failing, looks like a race condition:
[trace.zip](https://github.com/user-attachments/files/17528735/trace.zip)

```
1) openai.test.ts:56:2 › OpenAI Integration › models list is sorted correctly ────────────────────

    Test timeout of 5000ms exceeded.

    Error: expect(locator).toHaveText(expected)

    Locator: locator('div[role="option"]').first()
    Expected string: "gpt-3.5-turbo"
    Received string: " codegemma:latest 9B  "
    Call log:
      - expect.toHaveText with timeout 5000ms
      - waiting for locator('div[role="option"]').first()
      -   locator resolved to <div role="option" id="yfbQhZM0We" aria-selected="fal…>…</div>
      -   unexpected value " codegemma:latest 9B  "
      -   locator resolved to <div role="option" id="yfbQhZM0We" aria-selected="fal…>…</div>
      -   unexpected value " codegemma:latest 9B  "
      -   locator resolved to <div role="option" id="yfbQhZM0We" aria-selected="fal…>…</div>
      -   unexpected value " codegemma:latest 9B  "
      -   locator resolved to <div role="option" id="yfbQhZM0We" aria-selected="fal…>…</div>
      -   unexpected value " codegemma:latest 9B  "
      -   locator resolved to <div role="option" id="yfbQhZM0We" aria-selected="fal…>…</div>
      -   unexpected value " codegemma:latest 9B  "
      -   locator resolved to <div role="option" id="yfbQhZM0We" aria-selected="fal…>…</div>
      -   unexpected value " codegemma:latest 9B  "
      -   locator resolved to <div role="option" id="yfbQhZM0We" aria-selected="fal…>…</div>
      -   unexpected value " codegemma:latest 9B  "
      -   locator resolved to <div role="option" id="yfbQhZM0We" aria-selected="fal…>…</div>
      -   unexpected value " codegemma:latest 9B  "


      61 |
      62 | 		const modelOptions = page.locator('div[role="option"]');
    > 63 | 		await expect(modelOptions.nth(0)).toHaveText('gpt-3.5-turbo');
         | 		                                  ^
      64 | 		await expect(modelOptions.nth(1)).toHaveText('gpt-4');
      65 | 	});
      66 |

        at /Users/winston/projects/hollama/tests/openai.test.ts:63:37
```

---

This next test actually broke with the changes we made to Sessions.
[trace.zip](https://github.com/user-attachments/files/17528738/trace.zip)

I suspect that when you were running the tests locally you didn't have an Ollama server running at the same time, if you do you'll see the error I'm getting in the flash alert.

I think the issue is that we were mocking `**/api/tags` before and now we need to mock `**/api/chat`.

```
  2) session.test.ts:720:2 › Session › handles errors when fetching models ─────────────────────────

    Test timeout of 5000ms exceeded.

    Error: expect(locator).toBeVisible()

    Locator: locator('ol[data-sonner-toaster] li').filter({ hasText: 'Can\'t connect to Ollama server' })
    Expected: visible
    Received: hidden
    Call log:
      - expect.toBeVisible with timeout 5000ms
      - waiting for locator('ol[data-sonner-toaster] li').filter({ hasText: 'Can\'t connect to Ollama server' })


      741 | 		await expect(
      742 | 			page.locator('ol[data-sonner-toaster] li', { hasText: "Can't connect to Ollama server" })
    > 743 | 		).toBeVisible();
          | 		  ^
      744 | 	});
      745 |
      746 | 	test('can navigate out of session during completion', async ({ page }) => {

        at /Users/winston/projects/hollama/tests/session.test.ts:743:5
```

tests/openai.test.ts
```diff
@@ -0,0 +1,118 @@
...
+
+test.describe('OpenAI Integration', () => {
+	test.beforeEach(async ({ page }) => {
+		await page.goto('/settings');
+	});
+
+	test('fetches data from OpenAI with a correct API Key', async ({ page }) => {
+		await mockOpenAIModelsResponse(page, MOCK_OPENAI_MODELS);
+		await expect(page.getByText('Sync was successful')).toBeVisible();
+	});
```
fmaclen, line 19: Let's move `await expect(page.getByText('Sync was successful')).toBeVisible();` inside the `mockOpenAIModelsResponse()` function and remove this test.

fmaclen (approved):
This was certainly a tough one but I think we are good to go 👍 
Let's get it out!

PR #227, opened 2024-10-31
feat: Session: Use existing knowledge when submitting a prompt

GregoMac1:


GregoMac1:
The functionality is currently working but there are some things that I have to fix/improve.

fmaclen:
@GregoMac1 I just tried out the feature, here's a couple of notes:
- Once you select a Knowledge and submit the prompt, the Knowledge should become part of the session history and it should be removed from the `<PromptEditor>`, it could look something like this:
image
- The knowledge should have the option to delete it from the chat history (in the image above I used an `X` but you could use the trashcan icon).
- Maybe the knowledge should be added _before_ the prompt the user submits, that way the user can delete the context and then edit the prompt message to "re-trigger" the prompt without it.
- The combo boxes for Knowledge shouldn't have the `X`, they should be mandatory fields like the Language combobox in Settings. If the user doesn't want to add knowledge they will click on the trashcan icon to remove it.
- The auto-scroll behavior appears to have broken.

Once the core functionality is there I can give you a hand to get all of the layout stuff looking right, so don't stress too much about aligning things.

GregoMac1:
@fmaclen Thanks for the feedback! 

I wanted to ask you, should the knowledge be bounded to the session or to each specific message? For example, if I choose a knowledge and submit a prompt, should that knowledge be kept in the session to be used in the next messages? Or is it just for the original message?

fmaclen (changes requested):
I gave it a quick try and this is working great! 👍 

Left you some small comments. I'll try to cleanup the layout a bit tomorrow but I think this is more or less good to go. Let's get started with the tests.

fmaclen:
@GregoMac1 just pushed some minor changes to the design/layout. One note-worthy thing is I merged `AttachmentsToolbar.svelte` into `Prompt.svelte` because it felt like it was just adding unnecessary boilerplate code.

src/lib/sessions.ts
```diff
@@ -16,7 +16,7 @@ export interface Session {
 	id: string;
 	model: string;
 	messages: Message[];
-	systemPrompt: Message;
+	systemPrompt?: Message | Message[];
```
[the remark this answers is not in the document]
GregoMac1: Yeah, I did this so we can attach multiple knowledge (or images/urls in the future), but I know that it might not be the best solution. Do you have a better idea?

src/routes/sessions/[id]/Article.svelte
```diff
@@ -14,62 +14,87 @@
...
+					variant="outline"
+					href={generateNewUrl(Sitemap.KNOWLEDGE, message.knowledge.id)}
+					aria-label={$LL.goToKnowledge()}
+				>
+					{message.knowledge.name}
+					<Brain class="base-icon -mr-1 ml-2" />
+				</Button> -->
+			{:else if message.content}
+				<Markdown markdown={message.content} />
+			{/if}
```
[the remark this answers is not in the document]
Comment (GregoMac1) [reply src/routes/sessions/[id]/Article.svelte:94]: No, I think we don't. It's the previous implementation of knowledge in a message. I'll delete it.

src/lib/sessions.ts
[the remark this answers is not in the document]
fmaclen: Oh sorry, I missed this in my review.
You are submitting the "Knowledge" as a system prompt, I think we should submit it as a `role: 'user'` instead:

```json
[
  {
    "role": "user",
    "knowledge": {
      "id": "82bcfb",
      "name": "Susan",
      "content": "You are a helpful AI assistant made by Hollama.\nYour name is Susan.\nAnwer the user prompts the best you can.",
      "updatedAt": "2024-11-07T22:13:28.779Z"
    },
    "content": "You are a helpful AI assistant made by Hollama.\nYour name is Susan.\nAnwer the user prompts the best you can."
  },
  {
    "role": "user",
    "content": "hi, i'm steve"
  }
]
```

I think we can send multiple `user` messages at once but I don't think we can send multiple `system` prompts.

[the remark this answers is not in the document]
GregoMac1: Great then! I'll apply the necessary changes.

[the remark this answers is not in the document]
GregoMac1: Now that I think about it, do we need the `systemPrompt` at all?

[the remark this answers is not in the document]
fmaclen: We do, the `systemPrompt` it's a special type of message:

> System prompts are a special type of instruction used to set up the behavior of the model. It defines how the model should behave throughout the conversation. This prompt is usually provided at the start of the session to guide the tone, personality, and overall style of the AI’s responses.

src/routes/sessions/[id]/+page.svelte
```diff
@@ -221,13 +221,7 @@
...
 	{/if}
 
-	<PromptEditor
-		bind:model={$session.model}
-		{editor}
-		{handleSubmit}
-		{stopCompletion}
-		{scrollToBottom}
-	/>
+	<PromptEditor bind:session {editor} {handleSubmit} {stopCompletion} {scrollToBottom} />
```
Comment (fmaclen) [inline src/routes/sessions/[id]/+page.svelte:224]: At some point we renamed `PromptEditor` to `Prompt` but we didn't update the name.
Can you update the import and the name in on line 224?

```
import PromptEditor from './Prompt.svelte';
```

src/routes/sessions/[id]/Prompt.svelte
```diff
@@ -1,27 +1,45 @@
 <script lang="ts">
```
Comment (fmaclen) [inline src/routes/sessions/[id]/Prompt.svelte:1]: There are several type errors on this file.

src/routes/sessions/[id]/Controls.svelte
```diff
@@ -43,52 +36,9 @@
...
-					aria-label={$LL.newKnowledge()}
-					variant="outline"
-					href={generateNewUrl(Sitemap.KNOWLEDGE)}
-					class="h-full text-muted"
-				>
-					<Brain class="base-icon" />
-				</Button>
-			</svelte:fragment>
-		</FieldSelect>
-	</Fieldset>
```
Comment (fmaclen) [inline src/routes/sessions/[id]/Controls.svelte:90]: We don't want to remove the system prompt, we need to restore all of this code.

src/routes/sessions/[id]/Prompt.svelte
```diff
@@ -45,14 +63,42 @@
...
+
+	function submit() {
+		if ($attachments.length) {
+			const attachmentMessages: Message[] = [];
+			$attachments.forEach((a) => {
+				if (a.knowledge)
+					attachmentMessages.push({
+						role: 'user',
+						knowledge: a.knowledge,
+						content: a.knowledge.content
```
Comment (fmaclen) [inline src/routes/sessions/[id]/Prompt.svelte:87]: Let's add some formatting to the `content` value so the LLM understands what this text means

```jsx
`
CONTEXT
---
${a.knowledge.name}
---
${a.knowledge.content}
`
```

fmaclen (changes requested):
The feature is working but there are a few type errors and we need to restore the system prompt logic that was removed.

I believe our existing tests should have passed without modifications and right now there are several that are failing. Most of them will likely be fixed after restoring the system prompt stuff.

src/routes/sessions/[id]/Controls.svelte
Comment (GregoMac1) [reply src/routes/sessions/[id]/Controls.svelte]: You're right, I removed by mistake. I've just added it again.

src/routes/sessions/[id]/Prompt.svelte
Comment (GregoMac1) [reply src/routes/sessions/[id]/Prompt.svelte:1]: About this, I don't really know what the problem with the linter is. It seems like the `Knowledge` component is not being imported correctly.

However, the syntax looks fine (the component is imported from `Controls` as well and there isn't any problem there), the browser does not throw any errors, and if I run `npm run lint` it says that all files are okay.

I don't know how to get rid of that annoying lint error in the code editor.

src/routes/sessions/[id]/Prompt.svelte
Comment (GregoMac1) [reply src/routes/sessions/[id]/Prompt.svelte:87]: I'm sorry, I'm afraid I don't understand this syntax. I understand the context problem though, so where should we make the change?

Comment (fmaclen) [reply src/routes/sessions/[id]/Prompt.svelte:87]: Sorry, I meant that `content: a.knowledge.content` should be an interpolated string that includes the word `CONTEXT`, then the knowledge `name` and finally the `content` separated by `---`:

```typescript
attachmentMessages.push({
  role: 'user',
  knowledge: a.knowledge,
  content: `
CONTEXT
---
${a.knowledge.name}
---
${a.knowledge.content}
`})
```

Comment (GregoMac1) [reply src/routes/sessions/[id]/Prompt.svelte:87]: Thanks! All clear now.

src/routes/sessions/[id]/Prompt.svelte
```diff
@@ -1,27 +1,45 @@
...
-	export let model: string;
+	export let session: Writable<Session>;
 	export let handleSubmit: () => void;
 	export let stopCompletion: () => void;
 	export let scrollToBottom: (shouldForceScroll: boolean) => void;
 
 	let isOllama = false;
-	$: isOllama = $settingsStore.models?.find((m) => m.name === model)?.api === 'ollama';
+	let attachments: Writable<KnowledgeAttachment[]> = writable([]);
+	let knowledgeOptions: Writable<Knowledge[]> = writable([]);
```
Comment (fmaclen) [inline src/routes/sessions/[id]/Prompt.svelte:35]: I think the type errors we are because we are using a component as the type for the `let knowledgeOptions: Writable<Knowledge[]>` instead of an interface.

This will need to be renamed:

```svelte
import Knowledge from './Knowledge.svelte';
```

fmaclen (changes requested):
I found this bug where the **Knowledge** in the **Controls** page is conflicting with the `System prompt` field.

https://github.com/user-attachments/assets/97e34ede-f584-4e89-bf1d-f2f8f71d12ec

src/routes/sessions/[id]/Prompt.svelte
Comment (GregoMac1) [reply src/routes/sessions/[id]/Prompt.svelte:35]: Oh thank you! That was it. However, for some reason it wasn't enough with renaming the import, so I renamed `Knowledge.svelte` to `KnowledgeSelect.svelte` and now it works fine.

src/lib/components/FieldSelect.svelte
```diff
@@ -64,6 +64,7 @@
 	}
 
 	function handleOpenChange(wasMenuOpened: boolean) {
+		return;
```
GregoMac1, line 67: @fmaclen about this, what is the purpose of this function?

I am experiencing this bug, which I was able to fix by adding an early return to the `hanleOpenChange` function. It seems that when I select an option and I open the menu again but I don't change my selection, then the `FieldSelect` value is set to undefined, but the stored value remains the same.

https://github.com/user-attachments/assets/9bed30f2-ce18-4ec0-9196-0aae4af2bd7b

GregoMac1:
> I found this bug where the **Knowledge** in the **Controls** page is conflicting with the `System prompt` field.
> 
>  Area.mp4

@fmaclen Thank you for the bug report. I believe it's fixed now.

The only bug/unhandled behavior that I see now is when I add more knowledge attachments than those that fit on the screen. How do you think we can fix this?

https://github.com/user-attachments/assets/2ce6ce15-63fa-4b65-a8c7-33c3f3a905ad

src/lib/components/FieldSelect.svelte
fmaclen: I think the logic in `handleOpenChange` was there so that you could press `Escape` with the opened menu and it would clear the field. We have a bunch of assertions for this behavior (which are currently failing):

```typescript
// tests/ui.test.ts:79

// Click outside the combobox to apply the cleared value
await page.keyboard.press('Escape');
await expect(modelCombobox).toHaveValue('');
await expect(modelCombobox).toHaveAttribute('placeholder', 'Available models');
await expect(page.getByRole('option')).toHaveCount(0);
await expect(page.getByTitle('Clear')).not.toBeVisible();
```

I think the bug you are seeing is caused by this logic:

```typescript
$: if ($knowledgeStore)
	$knowledgeOptions = $knowledgeStore?.filter(
		(k) => !$attachments.find((a) => a.knowledge?.id === k.id)
	);
```

If instead of `$knowledgeOptions` we use the `$knowledgeStore` directly in the `KnowledgeSelect` everything works as expected, including highlighting the selected Knowledge:

image

Well, _"everything works"_ with the exception that now we don't filter out the "used" knowledge in the other fields, which definitely is a nice feature.

I think the solution would be to do something like this:

```svelte
<KnowledgeSelect
	value={attachment.knowledge?.id}
	options={$knowledgeStore?.filter((k) => 
		// Only filter out knowledge that's selected in OTHER attachments
		!$attachments.find((a) => 
			a.fieldId !== attachment.fieldId && // Skip current attachment
			a.knowledge?.id === k.id
		)
	)}
	showLabel={false}
	fieldId={`attachment-${attachment.fieldId}`}
	onChange={(knowledgeId) =>
		knowledgeId && handleSelectKnowledge(attachment.fieldId, knowledgeId)}
	allowClear={false}
/>
```

This way we can now get rid of `let knowledgeOptions: Writable<Knowledge[]> = writable([]);` completely.

src/lib/sessions.ts
```diff
@@ -112,6 +112,8 @@ export function formatSessionMetadata(session: Session) {
 }
 
 export function getSessionTitle(session: Session) {
-	const hasKnowledge = session.messages[0]?.knowledge;
-	return hasKnowledge ? session.messages[1]?.content : session.messages[0]?.content;
+	const firstUserMessage = session.messages.find(
+		(m) => m.role === 'user' && m.content && m.knowledge === undefined
```
fmaclen, line 116:
```suggestion
		(m) => m.role === 'user' && m.content && !m.knowledge
```

fmaclen (changes requested):
Left you some notes about the bug you were encountering.
Everything else looks good, let's get started with the test.

src/routes/sessions/[id]/Controls.svelte
```diff
@@ -1,19 +1,16 @@
...
 	import FieldInput from '$lib/components/FieldInput.svelte';
-	import FieldSelect from '$lib/components/FieldSelect.svelte';
 	import Fieldset from '$lib/components/Fieldset.svelte';
 	import P from '$lib/components/P.svelte';
 	import { loadKnowledge } from '$lib/knowledge';
 	import { knowledgeStore } from '$lib/localStorage';
 	import type { Session } from '$lib/sessions';
-	import { Sitemap } from '$lib/sitemap';
+
+	import Knowledge from './KnowledgeSelect.svelte';
```
Comment (fmaclen) [inline src/routes/sessions/[id]/Controls.svelte:13]:
```suggestion
	import KnowledgeSelect from './KnowledgeSelect.svelte';
```

src/routes/sessions/[id]/Controls.svelte
```diff
@@ -68,25 +65,12 @@
 <div class="controls">
 	<Fieldset>
 		<P><strong>{$LL.systemPrompt()}</strong></P>
-		<FieldSelect
-			label={$LL.knowledge()}
-			name="knowledge"
-			disabled={!$knowledgeStore.length}
-			placeholder={!$knowledgeStore.length ? $LL.emptyKnowledge() : ''}
-			options={$knowledgeStore?.map((k) => ({ value: k.id, label: k.name }))}
+		<Knowledge
```
Comment (fmaclen) [inline src/routes/sessions/[id]/Controls.svelte:68]:
```suggestion
		<KnowledgeSelect
```

src/routes/sessions/[id]/Prompt.svelte
```diff
@@ -109,39 +161,82 @@
 			</Field>
 		{/if}
 
+		{#if $attachments.length}
+			<div class="attachments">
+				{#each $attachments as attachment (attachment.fieldId)}
+					<div class="attachment">
+						<div class="attachment__knowledge">
+							<KnowledgeSelect
```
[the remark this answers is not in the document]
Comment (GregoMac1) [reply src/routes/sessions/[id]/Prompt.svelte:169]: Awesome! Thanks!

src/lib/components/FieldSelect.svelte
GregoMac1: Thank you! It worked like a charm. I removed `knowledgeOptions` and the code looks cleaner now.

GregoMac1:
@fmaclen I've added 4 test cases, please let me know if we should add any additional one.

fmaclen (changes requested):
Looks like we are missing test assertions for several bits of code.

The methodology I used was to comment out parts of the new logic we are introducing and was expecting one of the Knowledge tests to fail.

I added comments on the lines I commented out that will need tests or assertions.

GregoMac1:
Thanks for the feedback. I've added another test which I expect covers those three remaining scenarios.

fmaclen (approved):
Alright, I think we are good to go! 🚀

PR #235, opened 2024-11-16
feat: Session: Extract text from URLs

GregoMac1:


fmaclen:
This implementation will require significant amount of work and is no longer a priority.

PR #267, opened 2025-02-11
feat: add French language support

aginies:
Would be nice to have it also in French.


aginies:
> @aginies thanks for submitting the PR, this works so far but you are missing a few more changes:
> ....
> For reference, here's another PR for a recent language addition: #241

Thanks for the reference.
I fixed the missing part. Should be OK now .
regards.

fmaclen:
Thank you for making those extra changes 👍

PR #269, opened 2025-02-16
fix: Add .nvmrc to allow users to deploy on cloudflare pages

sainnhe:
This project can be hosted on cf pages, as the code in gh actions. The problem is that the node version used in cf pages is not 20 which will cause the deployment failing.

This PR adds a .nvmrc that specifies node version to 20, so cf pages can be successfully deployed.

To deploy it:

1. Fork this repo
2. In cf pages, create a new project from forked repo
3. Choose framework "SvelteKit"
4. Click deploy


fmaclen:
We are currently deploying  with an environment variable `NODE_VERSION=22.12.0`.
In other projects I used `.node-version` instead of `.nvmrc`, but we'll see if this works.

fmaclen:
Looks like it worked, thanks for the contribution.

PR #270, opened 2025-02-18
feat: add Arabic translation

aldoyh:
Hello there,

Just wanted to make it easier for other natives to interact with your amazing interface.

All the best.


README.md
fmaclen, line 1: The README.md needs to remain in English.

src/routes/settings/Interface.svelte
```diff
@@ -29,6 +29,7 @@
 		onChange={() => changeLanguage(value)}
 		options={[
 			{ value: 'en', label: 'English' },
+			{ value: 'ar', label: 'عربي' },
 			{ value: 'de', label: 'Deutsch' },
 			{ value: 'zh-cn', label: '中文 (简体)' },
 			{ value: 'es', label: 'Español' },
```
fmaclen, line 35: New languages should be added a the end of the options list.

fmaclen (changes requested):
Thanks for submitting this translation.
There are a few changes we need to make before rolling this out.

You are also missing a test, use this example from a [recent French translation](https://github.com/fmaclen/hollama/pull/267/files#diff-741b9c5e34d856c348b398d023013585492090694f186d67cb0a5fd67e6b0a77) we added as a reference.

fmaclen:
Also, make sure to run `npm run format` and then `npm run lint` to make sure the linter requirements are satisfied.

src/routes/settings/Interface.svelte
aldoyh: Definitely, I will do the required.
Thanks.

README.md
jhoermann: @fmaclen Maybe it would be interesting to have mutliple readme files for different languages - but that would be more of a Feature request.

aldoyh: Hi, actually I am with @jhoermann, how about we link them from the mother README.md file?

And yes @fmaclen, I will do it soon. Thanks.

jhoermann: > Github only shows one `README` file by default, keeping the `README` updated in multiple languages it's too much work so it's probably best if users just rely on Google Translate (or similar) for those.

See e.g. https://github.com/langgenius/dify you can just link the other readme files in the main readme. But as I said that's maybe not the highest prio.

fmaclen: I know you can link to other markdown files but I don't want to maintain multiple versions of the README every time we make changes to the main one. Please revert that change.

aldoyh: Sure thing dear,
But to answer that I'd say; Other languages should be the translators responsibility.

However, it's your call. Thank you. 💟

PR #272, opened 2025-02-24
feat: add Georgian translation

EkaterinePapava:


fmaclen (changes requested):
Thanks for submitting this translation.
Before we roll this out we need to make a few more changes:
- Add the language option to the menu / [Example](https://github.com/fmaclen/hollama/pull/267/files#diff-49d3f0b1c2a222ef8a54545a9f67d37e0ac01b4ce3a2c5687974ca18ad121617R35)
- Add a test / [Example](https://github.com/fmaclen/hollama/pull/267/files#diff-741b9c5e34d856c348b398d023013585492090694f186d67cb0a5fd67e6b0a77)

fmaclen:
Also, make sure to run `npm run format` and then `npm run lint` to make sure the linter requirements are satisfied.

EkaterinePapava:
Made both changes, also ran both commands, both seems to be satisfied

src/i18n/ka/index.ts
```diff
@@ -0,0 +1,163 @@
+import type { BaseTranslation } from '../i18n-types';
+
+const en = {
```
fmaclen, line 3: Should be `const ka`. Also update the default export below.

tests/locales.test.ts
```diff
@@ -128,4 +128,19 @@ test.describe('Locales', () => {
 			);
 		});
 	});
+	test.describe('Georgian', () => {
+		test.use({ locale: 'ka-GE' });
+		test('default language is Georgian', async ({ page }) => {
+			await page.goto('/settings');
+			expect(await page.evaluate(() => navigator.language)).toBe('ka-GE');
```
fmaclen, line 135: This test is not passing, I'm guessing it's related to the ISO code for the language but I'm not sure.
Can you double check?

tests/locales.test.ts
```diff
@@ -128,4 +128,19 @@ test.describe('Locales', () => {
...
+
+			await page.evaluate(() => window.localStorage.clear());
+			await page.reload();
+			await expect(page.getByText('Servers')).not.toBeVisible();
+			await expect(page.getByText('Serveurs')).toBeVisible();
```
fmaclen, line 140: This assertion is in French, should be in Georgian.

fmaclen (changes requested):
I ran the auto-formatter because there were still issues with the linter.
But at the moment the test isn't passing, check my comments below.

src/i18n/ka/index.ts
EkaterinePapava: fixed

tests/locales.test.ts
EkaterinePapava: ISO code for Georgian is ka-GE.

tests/locales.test.ts
EkaterinePapava: Looks better now?

EkaterinePapava:
Hello,
Is anything still needed from my side?

PR #293, opened 2025-04-02
fix: Reasoning is visible by default, auto-closes once the LLM completes the reasoning

### LINKED ISSUE #292: Suggestion: Default Expand Reasoning Process Display
Description:
Currently, the model's thought process is collapsed by default and not displayed. However, the majority of the reasoning steps are meaningful, and I find myself manually expanding them every time the model starts inferencing. I've observed that most web-based AI tools in the market default to showing the thought process expanded.

Proposed Change:
Could we consider changing the default behavior to expand the thought process? This would align with common practices in similar tools and reduce the need for repetitive manual actions.

Example:
https://chat.qwen.ai/
https://chat.deepseek.com/

rankaiyx:
Suggestion: Default Expand Reasoning Process Display

Closes #292

It seems like this boolean variable initialization should be modified, 
but I don't have a development environment and can't test it.


fmaclen:
I just pushed a small change that opens the _Reasoning_ box while it's reasoning, and then auto-closes when the answer stars. This matches the behavior I've seen in IDEs and several chat apps.

Still need to write a test for this feature but I'll try to get it out soon.

PR #335, opened 2025-10-10
fix: port mapping for VITE after container update

deenle:
Hi @fmaclen !
Thank you for the app and the latest container update!

I noticed that VITE is now explicitly exposed on port **3000**, causing the app not to map to port **4173** anymore.
After this update, I encountered an issue where I couldn't access the app through the previous mapping.

To resolve this, I changed the port mapping to **4173:3000**, which fixed the issue.

Best regards,
Tim


fmaclen:
Thanks for reporting this!

The issue is related to the [PR I just merged](https://github.com/fmaclen/hollama/pull/334). We were relying on Vite's preview command before but now we are running the server with a [Node command directly](https://github.com/fmaclen/hollama/pull/334/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R43) so it's defaulting to `3000`.

I think it'd be less friction for existing users to simply re-build the docker image by specifying the port number we used to have and keep the docs the way they were.

If you want to give it a shot, can you update the `Dockerfile` with these changes?

```diff
 # Set runtime environment
 ENV PUBLIC_ADAPTER='docker-node'
+ENV PORT=4173

 # Copy only the built application from builder stage
 COPY --from=builder --chown=appuser:appgroup /app/build ./build

 # Switch to non-root user
 USER appuser

-# Expose port (SvelteKit default is 3000)
-EXPOSE 3000
+# Expose port (vite preview default)
+EXPOSE 4173
```

deenle:
@fmaclen 
Thanks for fast feedback!

My bad! I originally thought I should update the Dockerfile, but decided not to intervene there.

fmaclen:
@deenle thank you for the contribution, much appreciate it!

PR #336, opened 2025-10-10
fix: port mapping for VITE after container update - 2

deenle:
@fmaclen I’d like to apologize for the mistake in the merged PR #335 where the port configuration was added only to the part of `build environment`.
Currently, added it to the `runtime` as well as `ENV VITE_ALLOWED_HOSTS='localhost'`.

P.S. Strangely, it was working well until reboot. :open_mouth:


Dockerfile
fmaclen: I just realized these aren't needed during the build step, so we can remove them.

Dockerfile
```diff
@@ -30,6 +30,8 @@ WORKDIR /app
 
 # Set runtime environment
 ENV PUBLIC_ADAPTER='docker-node'
+ENV VITE_ALLOWED_HOSTS='localhost'
```
fmaclen: I don't think we need to set `VITE_ALLOWED_HOSTS` since at this stage we are no longer dealing with the Vite runtime. Double check to make sure, but it should be safe to remove it.

fmaclen (changes requested):
Thanks for bringing this up again.

I think the actual fix is just to move the `ENV PORT=4173` to line 34 so it's set during the runtime, and then we can remove the other envs.

deenle (commented):
Did the edits, should be working fine

fmaclen (approved):
I think this looks good. Thanks again!

